PR #23, opened 2017-02-22
Fixed bug where get_array would consume self

frjonsen:
See https://github.com/mehcode/config-rs/issues/22

Fixed the bug, and added a simple sanity-check to tests


mehcode:
@frjonsen Thanks for noticing this. Definitely a typo on my part.

Issue #32, opened 2017-06-18
Merging an invalid but existing and non-required file should return an error


Anthony25:
Hello,
When merging a non-required file, in my opinion an error should be triggered if the file exists but cannot be parsed (invalid syntax in yaml, for example).
Currently, the result is Ok(), even if the parsing has failed.

What is your opinion about this behavior?
Thanks!

mehcode:
@Anthony25 I'd call this a bug. 

https://github.com/mehcode/config-rs/blob/master/src/file/mod.rs#L98-L103

If you want to submit a PR to fix that I'll merge it. Otherwise,  I'll get to it when I find time for v0.6.

Anthony25:
Thanks,
Sure, will do.

PR #33, opened 2017-06-21
Parsing errors even for non required files

Anthony25:
Related to #32
Even if a configuration file is not required, returns an error if
the file exists but the parsing failed.


mehcode:
Thanks @Anthony25

PR #37, opened 2017-06-23
Make boxed errors Send + Sync

Jascha-N:
This PR makes `ConfigError` play nicely with crates such as `error-chain` which require an error type to be `Send` by changing all `Box<Error>` into `Box<Error + Send + Sync>` and thereby making `ConfigError: Sync + Send`. The `Sync` is there because the standard library only implements:

```rust
impl<'a, E: Error + 'a> From<E> for Box<Error + 'a> {}
impl<'a, E: Error + Send + Sync + 'a> From<E> for Box<Error + Send + Sync + 'a> {}
```

This should not be a problem though, since almost all error types implement both `Sync` and `Send`, including the ones used by this crate.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -50,7 +50,7 @@ pub enum ConfigError {
 
         /// The captured error from attempting to parse the file in its desired format.
         /// This is the actual error object from the library used for the parsing.
-        cause: Box<Error>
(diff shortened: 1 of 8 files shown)


mehcode:
Thank you @Jascha-N

PR #44, opened 2017-09-01
Added HJSON (Human-Readable JSON) as a config file format

raphaelcohn:
HJSON is, for all intents and purposes, an alternative serialization of JSON.

This pull request adds support for it. It reuses the excellent work done by @laktak in [hjson-rust](https://github.com/hjson/hjson-rust) and integrates with serde.

(Background: I'm working on a project where TOML is too limited in its support of enums but JSON is too unpleasant to read for lots of long strings).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,9 +14,10 @@ license = "MIT/Apache-2.0"
 travis-ci = { repository = "mehcode/config-rs" }
 
 [features]
-default = ["toml", "json", "yaml"]
+default = ["toml", "json", "yaml", "hjson"]
 json = ["serde_json"]
 yaml = ["yaml-rust"]
+hjson = ["serde-hjson"]
 
 [dependencies]
 lazy_static = "0.2"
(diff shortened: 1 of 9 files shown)


Cargo.toml
```diff
@@ -14,9 +14,10 @@ license = "MIT/Apache-2.0"
 travis-ci = { repository = "mehcode/config-rs" }
 
 [features]
-default = ["toml", "json", "yaml"]
+default = ["toml", "json", "yaml", "hjson"]
```
mehcode: I'd rather we not add this to `default`. I'm still mulling over if `default` should be empty or not.

raphaelcohn: I'm not sure this decision is part of this PR. At the moment, the docs implied when I read them that neither json or yaml is a default choice.

I'd suggest that formats, in the long run, are broken out into separate crates, a la serde.

mehcode:
Agreed. Thanks for contribution!

PR #46, opened 2017-10-07
Fix markdown lists

svenstaro:
The lists were wrongly formatted. In markdown, the lists needs a free line before and after them.


mehcode:
Thanks for the fix @svenstaro :+1:

PR #54, opened 2017-11-14
Add derive(PartialEq) for Value, ValueType

cmyr:
This would be very helpful for me in diffing collections of configs.


mehcode:
Good addition. Thanks.

PR #56, opened 2017-12-14
remove unnecessary env key manipulation

### LINKED ISSUE #47: Picking up values from environment not working
Hi,

I am trying to use the environment only, to grep the configuration. Because that did not work, I also tried with an additional Settings.toml. Then the value from file is picked up, but not the environment.

Code:

    let mut settings = config::Config::default();
    settings
       .merge(config::File::with_name("Settings"))
        .unwrap()
        .merge(config::Environment::with_prefix("APP"))
        .unwrap();

    let x = settings.get::<String>("foo").expect("ok");
    println!("CFG: {:?}", x);

Settings.toml:
`foo = "20"`

Calling the program with `APP_FOO=10 target/debug/prog`. Only the value from toml is picked up.

I must be doing something very stupid, but I cannot get this to work. Can you spot my error?

**UPDATE** Also tried the simple example, with APP_DEBUG=FOO without success. The toml file wins.

The crazy thing is that to me the code in Environment looks fine, I am just not getting what is going on.

**UPDATE**
See new comment https://github.com/mehcode/config-rs/issues/47#issuecomment-339638438 below.

ChriFo:
I want to resolve #47 to get environment handling working again.

### DIFF
--- a/src/env.rs
+++ b/src/env.rs
@@ -85,9 +85,6 @@ impl Source for Environment {
                 }
             }
 
-            // Replace `separator` with `.`
-            key = key.replace(&self.separator, ".");
-
             m.insert(
                 key.to_lowercase(),
                 Value::new(Some(&uri), ValueKind::String(value)),


mehcode:
See https://github.com/mehcode/config-rs/issues/47#issuecomment-360838711

ChriFo:
Please show me how to override a key with underscore by setting an env variable.
For example **consumer_token** in this [example](https://github.com/mehcode/config-rs/blob/master/examples/hierarchical-env/src/settings.rs#L19).

mehcode:
You're right. It's not possible right now. But I don't want this approach to fix it. I explained an alternative fix in #47. If there is a PR submitted that does that, I'll accept it.

PR #50, opened 2017-10-27
use nom's error_kind() to be robust againt nom feature selection accross a project

kali:

PR #58, opened 2018-01-20
Be compatible with nom's "verbose-errors" feature

jethrogb:
See https://users.rust-lang.org/t/shared-dependencies-and-incompatible-features/15145

It's kind of hard to test if this really works since you'd need to run the test suite twice, once like normal, and once with `--features nom/verbose-errors`.

I also noticed the comments on the `to_result` conversion were incorrect with regards to what nom actually does. Not sure if you wanted the behavior specified, but that would require some different code. For now I just changed the comments.

### DIFF
--- a/src/path/parser.rs
+++ b/src/path/parser.rs
@@ -63,9 +63,9 @@ pub fn from_str(input: &str) -> Result<Expression, ErrorKind> {
                         expr = expr_;
                     }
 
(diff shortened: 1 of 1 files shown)


ChriFo:
Duplicate of #50

mehcode:
Merged in #50 so that should have taken care of this. Thanks anyway. 👍

PR #59, opened 2018-01-28
Adjust environment variable handling

### LINKED ISSUE #47: Picking up values from environment not working

ChriFo:
One more try to resolve #47

### DIFF
--- a/examples/hierarchical-env/src/settings.rs
+++ b/examples/hierarchical-env/src/settings.rs
@@ -53,7 +53,7 @@ impl Settings {
         // This file shouldn't be checked in to git
         s.merge(File::with_name("config/local").required(false))?;
 
-        // Add in settings from the environment (with a prefix of APP_)
+        // Add in settings from the environment (with a prefix of APP)
         // Eg.. `APP_DEBUG=1 ./target/app` would set the `debug` key
         s.merge(Environment::with_prefix("app"))?;
 
--- a/examples/hierarchical-env/src/settings.rs
+++ b/examples/hierarchical-env/src/settings.rs
@@ -53,9 +53,9 @@ impl Settings {
         // This file shouldn't be checked in to git
         s.merge(File::with_name("config/local").required(false))?;
 
-        // Add in settings from the environment (with a prefix of APP)
+        // Add in settings from the environment (with a prefix of APP_)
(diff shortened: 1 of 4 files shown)


mehcode:
Looks great! Only change I'd suggest is to make is to force a `_` after the env key prefix. What do you think? A similar library in Go, [viper](https://github.com/spf13/viper), does this.

theherk:
I know this is closed, but on the topic, the in-code samples suggest `with_prefix("APP")` but what actually works is `with_prefix("app")`. I'm happy to contribute to resolving this, but I'm unclear on the intended behavior. Should prefix be written as upper or lower, and based on code, should the environment be upper or lower. Then based on the env variable being upper or lower, should the value set be upper or lower.

I'm thinking the most simple magic here is to upper whatever is put into `with_prefix` so that "app", "aPp", and "APP" are all "APP_" in the environment, then lower everything once parsed so `app_kEy`, `aPP_key`, and `APP_KEY` are all represented by the symbol `key` in code.

ChriFo:
@theherk I added a test to demonstrate the behavior of this PR. As you can see lower, upper and even mixed cases for both - prefix and environment variables - are suitable!

mehcode:
This is great @ChriFo. Thanks for the contribution.

PR #78, opened 2018-09-26
Treat empty environment variables as unset

philbooth:
Not sure what you'll think of this change, but it comes from [a request from our ops folk](https://github.com/mozilla/fxa-email-service/issues/191). Essentially they were expecting an empty environment variable to fall back to the default value rather than overwrite it.

From an abundance of caution it's conditionally compiled behind a feature flag, as I presume it will be a breaking change for some people. Is there any chance of landing this (or something like it)?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,6 +19,7 @@ json = ["serde_json"]
 yaml = ["yaml-rust"]
 hjson = ["serde-hjson"]
 ini = ["rust-ini"]
+ignore-empty-env-vars = []
 
 [dependencies]
 lazy_static = "1.0"
--- a/src/env.rs
+++ b/src/env.rs
@@ -74,6 +74,11 @@ impl Source for Environment {
         };
 
         for (key, value) in env::vars() {
+            // Treat empty environment variables as unset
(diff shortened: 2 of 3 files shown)


mehcode:
Hm. As I wrote a large chunk of this lib from reading through [viper](https://github.com/spf13/viper/blob/master/viper.go#L970) it definitely feels like an oversight on my part as viper ignores empty env vars.

And then we have bash:

```
❯ FOO= echo ${FOO:-20}
20

❯ echo ${FOO:-20}
20
```

Technically it's a behavioral breaking change but I'd probably file this under a bug. I'll merge this in a slight variation so that I can remove the config later without a breaking change. You'll want to update to `0.9.1` and do `environment.ignore_empty(true)` here: https://github.com/mozilla/fxa-email-service/blob/73d07bd7cc89181f759f7cbc99be2687f9b88331/src/settings/mod.rs#L386

If/when we decide to make this the default/only behavior we can softly deprecate that method. Removing a feature flag is a breaking change.

philbooth:
Thanks @mehcode!

PR #79, opened 2018-10-03
update README: default features

vn971:
according to Cargo.toml, all features are now included by default,
hence a correction is needed.


mehcode:
Thanks for the correction.

PR #81, opened 2018-10-17
Update serde-hjson dependency

Ragnaroek:
serde-hjson update the outdated regex dependency in this version. This should make a cleaner dependency tree.


mehcode:
A patch change should come through with a `cargo update` without the need to change this project.

Thanks for the update though.

PR #85, opened 2018-11-10
support reading enums from config

geniusisme:
This fixes issue https://github.com/mehcode/config-rs/issues/74.

### DIFF
--- a/src/de.rs
+++ b/src/de.rs
@@ -5,7 +5,7 @@ use std::borrow::Cow;
 use std::collections::hash_map::Drain;
 use std::collections::HashMap;
 use std::iter::Peekable;
-use value::{Value, ValueKind, ValueWithKey};
+use value::{Value, ValueKind, ValueWithKey, Table};
 
 // TODO: Use a macro or some other magic to reduce the code duplication here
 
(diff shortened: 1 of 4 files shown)


src/de.rs
```diff
@@ -334,6 +359,107 @@ impl<'de> de::MapAccess<'de> for MapAccess {
...
+                _ => Err(self.structural_error()),
+            }?;
+            seed.deserialize(deserializer)?
+        };
+
+        Ok((value, self))
+    }
+}
+
+impl<'de> de::VariantAccess<'de> for EnumAccess {
```
geniusisme: Regarding panics in this impl:
- We already checked size of map, and one only can have VariantAcces from EnumAccess, which performs the check.
- If visitor chooses wrong variant to visit, leading to unreachable!(), then it is a buggy visitor, this is not a failure one can/should recover from.

geniusisme:
I guess we need to upgrade to serde-hjson v0.8.2 and update error check. Can do in this pull request.

mehcode:
Looks great. Thanks for this. Sorry for the delay in merging this. I'd love the update to hjson when you get time.

PR #87, opened 2018-11-30
upgrade serde-hjson to v0.8.2 so we can have consistent error check

geniusisme:
We had unfortunate situation where errors emitted by serde-hjson where different or v0.8.1 and v0.8.2. And we are expecting those errors in our test cases, so such a change is breaking for us. But as in semver terms minor version does not break, users could freely upgrade and then our tests failed (see travis).

That is the reasoning behind this version bump.


mehcode:
Ah. Makes sense. Thanks for looking into this.

Issue #83, opened 2018-10-24
Full path to broken field when Deserializing


vorner:
Hello

Let's say I want to deserialize some configuration using serde and something is broken (eg. missing field):

```rust
use config::{Config, File, FileFormat};
use serde_derive::Deserialize;

#[derive(Debug, Deserialize)]
struct Inner {
    test: String,
}

#[derive(Debug, Deserialize)]
struct Outer {
    inner: Inner,
}

const CFG: &str = r#"
inner:
    stuff: ABC
"#;

fn main() {
    let mut cfg = Config::new();
    cfg.merge(File::from_str(CFG, FileFormat::Yaml)).unwrap();
    let cfg: Outer = cfg.try_into().unwrap();
    println!("{:?}", cfg);
}
```

This produces this: `thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: missing field `test`', libcore/result.rs:1009:5`.

But if the configuration is bigger and the user forgot to put some config field in there, this will not help them to know in which section of configuration it is missing. Could this be added?

I'm guessing that `Config` is its own deserialization „format“ for serde and therefore has its own `de::Error` type. Also, it seems formats are able to put the path in somehow, eg. I can get something like this from serde-yaml:

```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Message("missing field `test`", Some(Pos { marker: Marker { index: 17, line: 3, col: 9 }, path: "inner" }))', libcore/result.rs:1009:5
```

(I'd much prefer to have the path actually in the message itself, because it is what eventually gets printed to logs or somewhere).

Dessix:
For anyone still running into this issue, https://crates.io/crates/serde_path_to_error worked well for this in my experimentation just now, thanks to `Config` implementing `serde::de::Deserializer`.
It's not perfect, but it's a lot better than having to manually walk the tree and attempt to deserialize manually. Sadly, the lack of an implementation for `&Config` means the operation is still destructive.

mattiecnvr:
When I use `serde_path_to_error` I am just getting an empty path every time. I think it is because our config has multiple sources, some of which are prefixed envs and some are files. Curious if anyone else experiences the same?

PR #89, opened 2018-12-30
Error path

vorner:
This pull requests tries to do something about #83. Unfortunately, that one *specific* case isn't solved by it, because that error message comes from within serde. Suggestions how to do something about it are welcome.

Though, this might help in some other cases.

Also, it cleans up few little things in the `de.rs`.

### DIFF
--- a/src/de.rs
+++ b/src/de.rs
@@ -2,6 +2,7 @@ use config::Config;
 use error::*;
 use serde::de;
 use std::collections::{HashMap, VecDeque};
+use std::iter::Enumerate;
 use value::{Value, ValueKind, ValueWithKey, Table};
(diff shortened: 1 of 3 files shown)


mehcode:
@vorner This is quite nice. Thank you. I'll merge this because it does improve the current state.

---

I'm in the middle of a total re-architecture of this library. One of several things being that removing source tracking from the happy path and using `DeserializeSeed` to pull out an error for a specific path from a specific file in the case of the error path. I really haven't had time to put a ton of work into it but I'll see about pushing it up as a branch. It's very WIP. There a number of things I'm very unhappy with in this library as it currently is.

Issue #60, opened 2018-01-31
When no configs are present, Config::new() should default to the structs default


elliott-davis:
In an application, sometimes there is no config present and you would just like to use the default values for your struct. If this is the case currently, you get an obtuse error like: `invalid type: unit value, expected struct Settings`. As an example usage:

```
#[derive(Debug, Deserialize)]
pub struct Settings {
    db_host: string,
}

impl Default for Settings {
    fn default() -> Self {
        Settings {
            db_host: IpAddr::V4(Ipv4Addr::new(127, 0, 0, 1)),
        }
    }
}

impl Settings {
    pub fn new() -> Result<Self, ConfigError> {
        let mut s = Config::new();
       // hand waiving pseudo code
        if config_file.present() {
           s.merge(File::with_name(config_file))?;
       }
       s.try_into()
}
```

Michael-F-Bryan:
I was looking for this exact feature. Originally I thought you could work around it by converting/serialising your `Settings` into a `Value` and swapping out the `Config`'s `cache` field. But there doesn't seem to be any way to do this...

elliott-davis:
@Michael-F-Bryan that sort of worked. However whenever I merged a file my cache override got blown away. After reading through the config merge code I see that the cache is repopulated from `ConfigKind` on each `refresh`.

Michael-F-Bryan:
I think a better solution would be to have a `with_default()` method which accepts some default object and if it's a map it'll iterate over the items, calling `set_default()` for each.

mehcode:
How about:

```rust
s.set_defaults::<Settings>()?;
```

`Config::set_defaults<T>` accepts a `T` that is `Serialize + Deserialize + Default`. Config could transform a generic `T` that way and set that as default values.

vorner:
I think there are two issues here, aren't there?

First one is how to use defaults ‒ and the answer for that one is to use the serde's default handling (eg. the `default` attribute on fields).

The other problem is the annoying `unit value` error. I managed to work around that one by merging an empty source as the first one, eg `config.merge(File::from_str("", FileFormat::Toml))?`. Could something like this be already done as part of the constructor? I mean, make sure whatever is tried to be deserialized looks like an empty map to serde, not unit.

mehcode:
I didn't see that before. Thanks for making that clear. This is more of a bug then. Marking to get fixed for 0.10.

Kirguir:
Creating a config so
```
[...]

PR #95, opened 2019-01-31
fix for issue #60

tanujitghosh:
Hi,

I'm new to rust programming and learning it during my free time, wanted to contribute and learn at the same time. 

I liked your library and wanted to help out
Worked on this issue #60 below. 
Please review and provide your feedback

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -11,6 +11,7 @@ use source::Source;
 
 use path;
 use value::{Value, ValueKind, ValueWithKey};
+use file::{File, FileFormat};
 
 #[derive(Clone, Debug)]
 enum ConfigKind {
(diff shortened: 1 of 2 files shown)


derekdreery:
Good work! You're a natural at rust.

One thing I would say is that hard-coding the workaround into the source is both inefficient (probably doesn't matter in this case as config parsing isn't performance sensitive), and makes the code more difficult to read. It's better if we can restructure the code so that the logic is correct. If we keep adding work-arounds into the source code we end up with code nobody can understand.

mehcode:
:bowing_man: Thank you for your contribution. I apologize for the late response. Moving forward I want to re-think the design of `config` as a whole. See #111 for more details.

---

I think ideally we want to have defaults be expressed as either:
 * An earlier layer in the configuration stack
 * Serde `#[serde(default)]` on the container level (which would delegate to the `Default` impl for the container)

The latter would work here if not for some strangeness with how the deserializer interacts with serde.

Issue #65, opened 2018-03-12
Fields with camelCase doesn't work


vjousse:
Hi there,
I'm new to rust and config-rs, and I'm facing an annoying problem: I can't get structs with `#[serde(rename_all = "camelCase")]` to work.
I've setup a [simple demo repository here](https://github.com/vjousse/test-config-rs) to illustrate my problem.
My Toml:
```toml
snakeTest = "foo"
```
My struct:
```rust
#[derive(Debug, Deserialize)]
#[serde(rename_all = "camelCase")]
struct Settings {
    snake_test: String,
}
```
My main:
```rust
fn main() {
    let mut c = config::Config::default();
    c.merge(config::File::new("Settings", config::FileFormat::Toml))
        .unwrap();

    // Deserialize the entire file as single struct
    let s: Settings = c.try_into::<Settings>().unwrap();

    println!("{:?}", s);
}
```
The error:
```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: missing field `snakeTest`', libcore/result.rs:916:5
```

Am I missing something? Thanks!

vjousse:
This seems to be related to this `to_lowercase` call here: https://github.com/mehcode/config-rs/blob/master/src/file/format/toml.rs#L32

Any reason why your are lowercasing every key?

mehcode:
The primary reason is it's not always straightforward what case the key is in when it comes into the application. Consider building your application targeting a configuration file where all keys are `lower_case` but the system administrator needs to override something with an environment variable as `UPPER_CASE`.

I'd be hesitant to start doing `camelCase` to `camel_case` conversions everywhere. It might make sense to allow uniformly changing keys from a specific source though. Perhaps something like this:

```rust
config.merge(
  File::new("Settings").key_replacer(|key| to_lower_case(key))
)?;
```

vjousse:
IMHO, a first good step would be adding some documentation about that. For example, warning people that keys in the files should be lowercase when using automatic deserialization with `try_into`.
What's your guidelines regarding the main `README.md`? It seems pretty poor atm, and this is very confusing for the newcomers. I may contribute some documentation if you want. Just tell me if I can help in that area.
Anyway, thanks for the answer and the great work!

bowbahdoe:
This is surprising behaviour to say the least. It makes reading in json written in camel case pretty hard and it seems like an unnessicary restriction.

mehcode:
[...]

PR #98, opened 2019-02-10
Minor simplifications in env

damiencollard:


mehcode:

---

Moving forward I think it makes sense to switch to using the `caseless` crate so we have case-insensitivity with case-preservation (would fix #65 among other issues).

PR #103, opened 2019-03-08
Added HOCON as a config file format

mockersf:
Default features of the hocon library are disabled.

The two features disabled are:
* serde support, which this crate already does
* `include url(...)` hocon syntax support, which needs a dependency on `reqwest` even though it won't be used in most cases


mehcode:

---

Going to close because how we add file formats is going to change pretty drastically. There is a lot of duplicated code in the different formats.

Thank you again though. 

As a future point we can probably support a feature flag like `hocon_url` to turn on `include` support.

Issue #111, opened 2019-05-07
Exploring a redesign and discussing how a modern Rust configuration library should behave


mehcode:
This library has gotten quite stale over the last year and I've been loathe to come back to it mostly because I don't like it. It's complex to use, complex to develop, etc. I apologize for letting it atrophy.

I'd like to thank you everyone who has contributed or submitted a PR ( especially those who've done so in the last several months and have just sat there ). It's not that I don't want to merge in your work, it's that if I come back to this I'll want to do too much and I haven't had the time available to do that.

---

With that out of the way I want to explore how a configuration library could behave in Rust, optimally, using some example use cases as a jumping off point:

```rust
// load configuration only from envrionment
let settings: Settings = config::env::vars()
    // prefix(strip, s) 
    // filter to only keys that case-insentively start with s
    // optionally strip that prefix
    // calling prefix more than once applies _more_ prefixes to search
    .prefix(true, "app_")
    // explictly transform key names to allow traversing nested configuration
    // replaces .separator("__") from config v0.9
    .map(|key| key.replace("__", "."))
    .load()?;

// find a single configuration file from example.* in
let settings: Settings = config::file::find("example")
    // look in the following paths (by default cwd)
    .paths(&[xdg_config_home, home])
    // expect format in TOML only
    // default allows all formats compiled in
    .format(Format::Toml)
    // and load into Settings
    .load()?;

// find multiple configuration files and merge together
let settings: Settings = config::file::find_all(&[
    "default",
    mode // development, production, or test, etc.
]).load()?;

// load configuration from arguments
let settings: Settings = config::env::args(clap_matches).load();

// manually create a configuration stack
let settings: Settings = config::Stack::new()
    // start with an embedded JSON file 
    .add(config::file::from_bytes(Format::Json, include_bytes!("default.json")))
    // override from environment
    .add(config::env::vars())
    // load into settings
    .load()?;

// create a dynamic configuration (ala config v0.9)
// any _source_ could be built into a Config struct
let settings: config::Config = config::file::find("example").build();

// a dyn configuration can listen for changes on supported sources
let _handle = settings.watch(|| do_something_fun_when_config_changes);

// use path access
let host: &str = settings.get("db.host");
```

---

Thoughts? How would _you_ want `config` to work?

mockersf:
Just a comment describing how I tend to setup configuration:
- Start with default values for config
- Then try to read a config file in XDG config folder and override
[...]

PR #112, opened 2019-05-13
Make from_toml_value case sensitive

epilys:
TOML is case sensitive [0], so key values should not be lowercased.

[0] https://github.com/toml-lang/toml#user-content-spec

### DIFF
--- a/src/file/format/toml.rs
+++ b/src/file/format/toml.rs
@@ -29,7 +29,7 @@ fn from_toml_value(uri: Option<&String>, value: &toml::Value) -> Value {
             let mut m = HashMap::new();
 
             for (key, value) in table {
-                m.insert(key.to_lowercase().clone(), from_toml_value(uri, value));
+                m.insert(key.clone(), from_toml_value(uri, value));
             }
 
             Value::new(uri, m)


mehcode:
:bowing_man: Thank you for your contribution. However, I do not wish to accept any breaking changes to this repository at this time. 

---

Note that I fully agree with changing this but it will be a part of #111.

epilys:
> Note that I fully agree with changing this 

I don't know if it's a matter of agreeing or not, since it's in the specification and now config-rs isn't compliant. It's a bug.

But of course it's your decision and I respect it.

mehcode:
@epilys Let me try and explain my reasoning here as its not _completely_ a bug. 

`config` has been about merging different sources of configuration with the most common example being a file and the environment. 

To make this merging process "easy", all configuration keys were lowercased when they entered the configuration store.

In hindsight that was not ideal. However, I can't remove the lowercasing without allowing for case-insensitive merging of configuration keys which is now depended on by several projects (mine included). The current solution for this is that in v0.10 configuration keys will be stored as received (no lowercasing) but all comparison operations will be done using case folding.

Issue #74, opened 2018-07-06
Support Enum?


quake:
Got this error when try it with config-rs:
```
Err(invalid type: string "master", expected enum Mode)
```

Below is the struct and toml file:
```
#[derive(Debug, Deserialize)]
struct Database {
    url: String,
    mode: Mode,
}

#[derive(Debug, Deserialize)]
enum Mode {
    Master,
    Slave,
}
```

```
[database]
url = "postgres://postgres@localhost"
mode = "master"
```

toml supports Enum
https://github.com/alexcrichton/toml-rs/pull/165

am I doing anything wrong?

mehcode:
It just seems that we don't support it on our end probably due to oversight. This would be a bug that needs fixing.

vorner:
I've hit this too and tried to report it to serde (because to me it looked like the generated Deserialize must have been wrong).

Anyway, I also have a workaround, look [here](https://github.com/serde-rs/serde/issues/1415) ‒ the trick with flattening another structure in. I don't know why it helps yet, though.

mehcode:
Huh. The code path for flatten might be strange. It's definitely our fault for the normal case though. Enums are handled different in the deserializer and we just are missing that.

vorner:
I just wanted to ask, if the code is in master now, would it make sense to release it as part of 0.9.2, not waiting for full 0.10.0?

mehcode:
Released as 0.9.2

vorner:
This seems not to work fully yet, this still fails :-( :

```rust
use std::collections::HashMap;
use config::{Config, File, FileFormat};
use serde::Deserialize;

#[derive(Debug, Eq, PartialEq, Hash, Deserialize)]
enum X {
    A,
    B,
}

const CFG: &str = r#"
A = "hello"
B = "world"
"#;

fn main() {
    let mut config = Config::new();
    config.merge(File::from_str(CFG, FileFormat::Toml)).unwrap();
    let c: HashMap<X, String> = config.try_into().unwrap();
}
```

```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: invalid type: string "b", expected enum X', src/libcore/result.rs:999:5
```

By the way… should I open a separate issue for the fact it tries to feed it with lower-case `b` instead of the `B` I put into the configuration? (but it still fails even if I rename the enum to accept lower-case version).

It does seem to work in value-position in a struct, though.

Can we reopen? Or should I open a new one?

geniusisme:
[...]

PR #118, opened 2019-08-05
Enum key

vorner:
This implements the enums as map keys from #74. This probably works only for simple (data-less) enums.

I hope this has no ill side effect to some other use case.


vorner:
I've noticed numeric keys (eg. `HashMap<usize, usize>`) doesn't work either. But I've discovered a slightly hacky way to make them work with just deserializing the key out of `Value` created from the string. I'll submit alternative merge request.

PR #124, opened 2019-11-19
feat: add optional toml feature to only depend on toml crate

rottedfrog:
I'd like to be able to include only the toml feature, but it's the only one that doesn't have a separate feature.


rottedfrog:
Actually not required - missed a key point of optional dependencies automatically being features...

PR #128, opened 2020-01-02
Apply clippy suggestions

0x1793d1:
Apply clippy suggestions:
- Redundant field names in struct initialization
- Use of deprecated item 'try': use the `?` operator instead
- Trait objects without an explicit `dyn` are deprecated
- Unused variable
- Writing `&String` instead of `&str` involves a new object where a slice will do.
- Use of `ok_or` followed by a function call
- This argument (1 byte) is passed by reference, but would be more efficient if passed by value (limit: 8 byte)
- Redundant clone
- You seem to be trying to use match for destructuring a single pattern. Consider using `if let`
- Identical conversion

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -18,7 +18,7 @@ enum ConfigKind {
     Mutable {
         defaults: HashMap<path::Expression, Value>,
         overrides: HashMap<path::Expression, Value>,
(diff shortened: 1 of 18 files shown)


jbabyhacker:
@0x1793d1 Why was this closed and not merged?

PR #129, opened 2020-01-13
Adding support for automatically injecting environment variables in configuration file

dawid-nowak:
Rationale;

Creating a different configuration file for different deployments could be tedious especially in cloud environment and we could end up with many different configuration files. It could be potentially easier to set environment variables and inject those into placeholders in the configuration file. 
The approach is similar to Spring application.properties where environment variables are automatically resolved and injected on deployment. 

This should tie in well with Docker and Kubernetes which make it easy to pass variables into containers.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,17 +19,19 @@ json = ["serde_json"]
 yaml = ["yaml-rust"]
 hjson = ["serde-hjson"]
 ini = ["rust-ini"]
+with_env_vars = ["regex"]
 
 [dependencies]
 lazy_static = "1.0"
 serde = "1.0.8"
 nom = "5.0.0"
-
 toml = { version = "0.5", optional = true }
 serde_json = { version = "1.0.2", optional = true }
 yaml-rust = { version = "0.4", optional = true }
 serde-hjson = { version = "0.9", optional = true }
 rust-ini = { version = "0.13", optional = true }
+regex = {version="1",optional=true}
+
 
 [dev-dependencies]
 serde_derive = "1.0.8"
--- a/README.md
+++ b/README.md
@@ -37,6 +37,16 @@ config = "0.9"
 See the [documentation](https://docs.rs/config) or [examples](https://github.com/mehcode/config-rs/tree/master/examples) for
 more usage information.
 
+
+### Environment Variables Injection
+ - `with_env_vars` - adds support for environment variable injection into the configuration file. See [example](./examples/env_injection).
+
+
+
+
+
+
+
 ## License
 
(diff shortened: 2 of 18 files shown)


mehcode:
What benefit does this have over using the already existing environment variable merging?

See: https://github.com/mehcode/config-rs/blob/master/examples/simple/src/main.rs#L7

dawid-nowak:
@mehcode 
I think the key advantages are:
1. environment variables are not related to the key in the configuration file and because of that 
2. injecting variables into complex multil level config files is relatively straightforward

for example I am not sure how using merge option would work with following yaml file
client_ip: "0.0.0.0"
channels:
  kafka:
    - brokers: ${kafka_brokers}
      consumer_topics:
        - consumer_topic: ${consumer_topic_1}
          consumer_group: ${consumer_group_1}
          client_topic: ${client_consumer_topic_1}
      producer_topics:
        - producer_topic: ${producer_topic_1}
          client_topic: ${client_producer_topic_1}

  nats:
    - brokers:
        - "nats://docker_nats_1:4222"
      consumer_topics:
        - consumer_topic: "Response"
          consumer_group: "swir"
          client_topic: "SubscribeToAppB"
      producer_topics:
        - producer_topic: "Request"
          client_topic: "ProduceToAppB"

dawid-nowak:
@mehcode Any update on this ? Perhaps there is another way to deal with it so if you have any suggestions I will have a look into it.

geniusisme:
Wouldn't separate function be a better option?
Like `fn interpolate_env(File) -> File `

This allows finer-grained control (compile feature allows to interpolate all config files or non of them).  It also separates concepts of file and env variable, code-wise.

dawid-nowak:
@geniusisme 
sure; i don't mind doing it. 
could you elaborate more ? where do you want that function to be defined/used.

mehcode:
Have you looked into HOCON files? Sorry for the delay,  not a lot of time lately for this.

dawid-nowak:
@mehcode No worries
I presume you mean this https://github.com/lightbend/config/blob/master/HOCON.md 
I must admit I have not seen it before even though I used Play in the past. It seems that they do substitutions in a slightly different way to SpringBoot. 
So my questions are: 
1)  whether you want me to change the pull request to follow the guidelines from HOCON in terms of how we define substitutions with a fallback to environment variables as defined in that document. 
2) implement completely new handler for HOCON files ?

dawid-nowak:
I think i will close this pull request as I am no longer sure that this is a valid approach. While injecting variables like this might work for small configs or very selected use cases, based on my experiences with configuring Swir sidecar, this approach won't scale for more complex configuration files.

PR #133, opened 2020-03-14
Introduce an option for disabling the automatic resolver in FileSourceFile

eisterman:
A new feature with granted retro compatibility.
To solve the problem introduced in https://github.com/mehcode/config-rs/issues/132 I have introduced the solution number 3:

> Add a flag to the `FileSourceFile` struct named `use_resolver` and add a method in `File<FileSourceFile>` similar to `required(bool)` where the user can enable or disable the resolver in `FileSourceFile` (The best solution. No breaking change, no removing of feature, only adding an option flag that if not used will be set to `True` for retro compatibility)

The struct File now has two new methods, `with_exact_name` and `exact_name`, the first to create a File with a FileSourceFile imposed to not search the extension of the file.
With this approach, one can load a file named `Settings.wrongextension` as a Toml file like before, but if the file is missing the program will not search another file named for example `Settings.yaml` without asking.

I've introduced three tests for the new methods, and a very small fix of the wrong extension error message, because before it used the absolute path when the not found error used the relative/given path.

### DIFF
--- a/src/file/mod.rs
+++ b/src/file/mod.rs
@@ -55,6 +55,20 @@ impl File<source::file::FileSourceFile> {
             source: source::file::FileSourceFile::new(name.into()),
         }
     }
+
+    /// Given the full name of a file, will use it only if with the exact name without
+    /// any attempt to locate another file. It will analyze the extension if the property
+    /// format isn't setted but without using file with a different fullname.
+    pub fn with_exact_name(name: &str) -> Self {
+        Self::with_name(name).exact_name(true)
+    }
+
+    /// If enabeld, if a file with the exact name is not found,
+    /// will not attempt to locate a file based on the format property.
+    pub fn exact_name(mut self, flag: bool) -> Self {
+        self.source.disable_file_resolve(flag);
+        self
+    }
 }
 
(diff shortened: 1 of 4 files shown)


mehcode:
I see where this is coming from. First, thanks for the contribution. I agree as well that your solution #3 is probably the best way to handle this.

Very minor nit, should we have the option method simply `exact` instead of `exact_name`? I think it conveys the same information in context:

```rust
File::with_name("Config.toml").exact(true)
File::with_exact_name("Config.toml")
```

eisterman:
I agree it is certainly more coherent and sensible as a name.
I sent the name change commit.

eisterman:
Do you need anything else for the merge?

eisterman:
Any problems?

matthiasbeyer (approved):

matthiasbeyer:
Are you still interested in this? 

If yes, I created a maintenance fork (read [here](https://www.reddit.com/r/rust/comments/m43p3p/configmaint_maintenance_fork_of_config/)).
Feel free to submit your patches! :smile:

matthiasbeyer:
Can you rebase this to latest master?

PR #151, opened 2020-09-06
Fix paths in tests expectations for all platforms

szarykott:
This is first thing I came across while trying to do some changes in code. I started tests and they failed due to different way paths are structured on Linux and on Windows, which is my development platform.

Here I did some changes so that expectations should work correctly on all the platforms.

### DIFF
--- a/tests/errors.rs
+++ b/tests/errors.rs
@@ -26,7 +26,7 @@ fn test_error_parse() {
     assert!(res.is_err());
     assert_eq!(
         res.unwrap_err().to_string(),
-        format!("failed to parse datetime for key `error` at line 2 column 9 in {}", path.to_str().unwrap())
+        format!("failed to parse datetime for key `error` at line 2 column 9 in {}", path.display())
     );
 }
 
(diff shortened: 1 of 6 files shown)


tests/errors.rs
```diff
@@ -20,10 +21,12 @@ fn test_error_parse() {
     let mut c = Config::default();
     let res = c.merge(File::new("tests/Settings-invalid", FileFormat::Toml));
 
+    let path : PathBuf = ["tests", "Settings-invalid.toml"].iter().collect();
+
     assert!(res.is_err());
     assert_eq!(
         res.unwrap_err().to_string(),
-        "failed to parse datetime for key `error` at line 2 column 9 in tests/Settings-invalid.toml".to_string()
+        format!("failed to parse datetime for key `error` at line 2 column 9 in {}", path.to_str().unwrap())
```
joelgallant, line 29: nit: paths have a `.display()` to avoid unwraps like this

szarykott: Thank you for the hint!

joelgallant (approved):
Thanks! It would be good to add windows CI to prevent errors like this.

PR #157, opened 2020-10-18
Replace travis with Github actions

Nukesor:
This commit replaces the travis CI with Github's new actions.

It's just a proposal, however it's integrations are pretty nice and one doesn't need to visit another slow website to check which branch of the CI actually failed.

I tried to fix the currently failing tests, however those seem to be upstream in the hjson crate?
I'm not exactly sure what's happening though.

The new CI also ensures, that everything compiles on MacOS and Windows.
If this isn't something that's needed or wanted for this crate, I'll just remove it :)


MarcoIeni:
you should add 
```
        if: matrix.os == 'ubuntu-latest'
```
also to rustfmt, since there is no need to check format all the time.

In addition, maybe you want to check rustfmt and clippy only on `stable` toolchain?

Nukesor:
@MarcoIeni You're right. I did that for my own projects, but completely forgot about this PR :sweat_smile:
Commit has been adjusted

matthiasbeyer:
This is superseeded by #175

PR #164, opened 2021-01-29
A public ValueKind allows external Source implementations

brosander:


fpoli:
Are you sure `ValueKind` needs to be public? See [here](https://github.com/viperproject/prusti-dev/blob/22a4eb83ef91391d9a91e6b3246ddf951b8eb251/prusti-common/src/config/commandline.rs#L97) for an example of how to implement a `Source` in a different crate.

brosander:
Good call, I'd been going off the env implementation but it's not necessary

https://github.com/mehcode/config-rs/blob/master/src/env.rs#L114

Issue #163, opened 2021-01-06
What happens with relative pathes?


matthiasbeyer:
Hi.

I have a question about relative pathes in layered configuration. For example if I have the following structure:

```
/
/root.toml
/foo/foo.toml
/foo/bar/bar.toml
/foo/bar/somefile.txt
/baz/baz.toml
/baz/somefile.txt
```

And I layer my configuration objects via the nesting, so I get two configuration objects:

1. `root.toml + foo.toml + bar.toml`
2. `root.toml + baz.toml`

Where do I end up with a `path = ./somefile.txt` in `baz.toml`? The `PathBuf` deserialized does not point to the right file, does it?
And I cannot find it after deserializing the objects because 

1. I don't know which _file_ the setting came from, so I cannot search relative to that files directory
2. If I would search the tree, `foo/bar/somefile.txt` would be found first, which is obviously wrong

---

Is there a strategy to handle such cases? Is there a way to make `config-rs` handle relative pathes?
I think it is in the scope of this crate and would love to see such a feature, even contribute code for it.

matthiasbeyer:
Hi @mehcode any thoughts on this?

matthiasbeyer:
I'm still interested in this.

matthiasbeyer:
Hi.

This gets more and more relevant for me. I'm thinking about implementing it myself for this library, if it is okay with you to include such a change. I have to talk to my employer first, though, whether I can implement this on company time.
Still: Would you merge such a feature?

fpoli:
You could try contacting the other maintainers mentioned here: https://github.com/mehcode/config-rs/issues/149#issuecomment-686272557

matthiasbeyer:
Yeah, I'll do that.
I'll use the shotgun approach: @dawid-nowak @eisterman @ignatenkobrain @joelgallant @LukeMathWalker @szarykott @up9cloud @vmalloc

matthiasbeyer:
So, as the crate does not necessarily know about `PathBuf`s, we cannot handle them in the codebase of this crate. Though, I was able to solve my usecase via the newly introduced `Config::set_once()` function (if anyone is interested how, see [this](https://git.sr.ht/~matthiasbeyer/butido/tree/v0.1.0/item/src/repository/repository.rs#L70) scope, or if that is too complicated shoot me an email :smile: ).

I'll close this now.

PR #165, opened 2021-02-19
Add getter for getting a &mut Value from Config

matthiasbeyer:
This is a (maybe possible) solution for #163 

This is, of course, not as nice as it could be, because it exposes interior of this crate to the user.
I'm currently experimenting with these patches in my source, to see whether such a thing is feasible at all.

Feel free to ignore until I reached a state where it works, then we can debate over it.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -168,7 +168,7 @@ impl Config {
         }
     }
 
-    pub fn get_value_mut<'de, T: Deserialize<'de>>(&mut self, key: &str) -> Result<&mut Value> {
+    pub fn get_value_mut(&mut self, key: &str) -> Result<&mut Value> {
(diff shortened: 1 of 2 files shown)


matthiasbeyer:
Okay, so I was able to confirm that this is a possible way to implement the feature I need.

Here's how I use this feature now: https://git.beyermatthi.as/butido/commit/?h=pathes-to-absolute&id=370d6bc6e5140d5fa26c0c7029bc1269e9b0c9ad

I think we can start a discussion about this.

matthiasbeyer:
btw, it looks like the CI failure is not caused by me, is it?

matthiasbeyer:
Superseded and thus not relevant anymore.

PR #169, opened 2021-03-04
Fix test failure

### LINKED ISSUE #158: config-rs broken since rust 1.48
We're trying to use `config-rs` using the latest rust `nightly-1.49-0`, and we're getting:

```
thread 'main' panicked at 'attempted to leave type `linked_hash_map::Node<std::string::String, value::Value>` uninitialized
, which is invalid', /home/tyler/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/
mem/mod.rs:659:9
```

I see this is also failing on your travis build too:

https://travis-ci.org/github/mehcode/config-rs/jobs/734764347#L291

cc @Nutomic

matthiasbeyer:
As reported in

    https://github.com/hjson/hjson-rust/issues/23

the issue of the failing tests is the serde_hjson crate. The problem can
be prevented (as reported by
https://github.com/hjson/hjson-rust/issues/23#issuecomment-775520018) by
disabling the default features of the crate (namingly preservation of
key order).

This commit disables the default features of serde_hjson to fix our
tests.

--

Closes #158

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,7 +28,7 @@ nom = "5.0.0"
 toml = { version = "0.5", optional = true }
 serde_json = { version = "1.0.2", optional = true }
 yaml-rust = { version = "0.4", optional = true }
-serde-hjson = { version = "0.9", optional = true }
+serde-hjson = { version = "0.9", default-features = false, optional = true }
 rust-ini = { version = "0.13", optional = true }
 
 [dev-dependencies]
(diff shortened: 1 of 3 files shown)


matthiasbeyer:
This is merged in [config-rs-maint](] and) and released in [version 0.11.0](https://git.sr.ht/~matthiasbeyer/config-rs-maint/refs/v0.11.0).

PR #177, opened 2021-03-17
version 0.11.0

matthiasbeyer:
I'd like to make a release with the newly landed features.

AFAIK, I can publish on crates.io.


matthiasbeyer:
The changelog is best-effort as there is no 0.10.0 or 0.10.1 tag in the repo, yet. So I checked compiled the changelog by assuming the 0.10.1 version from the history.

Issue #167, opened 2021-02-23
Edgecase: Only signed integers (64 bit) are supported?


matthiasbeyer:
As far as I can see, only i64 are supported right now by this crate, am I right?

```
ValueKind::Integer(i64)
```

If someone wants to store a `u64`, they will run into an error, because we're casting `u64` to `i64`. This is most likely not a problem for any real-world configuration file, but should be fixed anyways.
Especially because #161 wants 128 bit integer support.

jaskij:
Since #178 is merged, shouldn't this be closed? Or is there something else that needs to be done?

matthiasbeyer:
Whoopsi, you are right! :tada: :laughing:

Issue #161, opened 2020-12-30
128-bit integers are not supported


nagisa:
Trying to `Config::get` a structure that contains u128 fields will fail with an error from serde.

> u128 is not supported

---

config-rs should override the u128 and i128 methods to support these integers.

matthiasbeyer:
I guess this means that `Deserializer::deserialize_u128` and similar functions must be implemented on `Value`, right? Also, the internal `ValueKind` type must be able to hold a 128 bit sized integer.

Would you like to implement this in a PR?

matthiasbeyer:
I just added a [patch](https://git.sr.ht/~matthiasbeyer/config-rs-maint/commit/116b860bc40e283518243139fe54ac5a0be8c6d6) to config-rs-maint (not in master yet) that adds support for 128bit integers (signed).

Would you like to contribute tests for this patch?

nagisa:
Hi, I'll see if I can make time to contribute some tests. Not super confident I'll be able to, though.

PR #178, opened 2021-03-17
Add support for different sized integers

matthiasbeyer:
This also enables support for 128 bit integers.
Nothing is tested, though.

---

I guess this is a draft for now.


matthiasbeyer:
This is related to #161 and #167

Nukesor:
Great to see you taking care of the project!

Cheers

matthiasbeyer:
Thanks :heart:

matthiasbeyer:
So, I guess this PR should add a bunch of tests to check

* Are the right integer types are used
* Do things fail properly if the size of the types in use are wrong
* Edgecases (`i8::MAX` vs `u8::MAX`) - any maybe we can add `u...` support as well
* possibly more

matthiasbeyer:
Hm, so after some more investigation: Our backend crates do not support integers smaller than i64. So I guess there's no point in supporting them either, is there?

matthiasbeyer:
@Nukesor and @nagisa - would you like to review my patchset here? There are too few tests, if you can provide some, that'd be awesome, too!

src/value.rs
```diff
@@ -241,6 +319,172 @@ impl Value {
         }
     }
 
+    /// Returns `self` into an i128, if possible.
+    pub fn into_int128(self) -> Result<i128> {
+        match self.kind {
+            ValueKind::I64(value) => Ok(value as i128),
```
nagisa, line 325: Consider `.into()` instead of casting.

src/value.rs
```diff
@@ -241,6 +319,172 @@ impl Value {
...
+            ValueKind::I128(value) => Ok(value),
+            ValueKind::U64(value) => Ok(value as i128),
+            ValueKind::U128(value) => Err(ConfigError::invalid_type(
+                self.origin,
+                Unexpected::U128(value),
+                "an signed 128 bit integer",
+            )),
+
+            ValueKind::String(ref s) => {
+                match s.to_lowercase().as_ref() {
```
nagisa, line 335: Why do we do this, but don't transparently allow, say, `u128` values that are in range of `i128` to convert to `i128`?

nagisa:
This broadly LGTM, but I have no say over this project, myself.

src/value.rs
matthiasbeyer: I guess you wanted another line to be commented here? I will investigate where we can do this, yes!

matthiasbeyer:
Still, your opinion matters to me. Thanks for the review!

src/value.rs
nagisa: No, I meant to ask why we transparently convert strings like `yes` or `no` to integers, but not integers to other integers. Seemed somewhat counterintuitive.

matthiasbeyer: Ah, now I see what you mean.

This patch tries to mimic the behavior of the non-int128 code... I agree that this might be not what we want for the next release, but I would introduce a change to that behavior in another PR, not in this one.

matthiasbeyer:
Rebased on master. Still not sure whether I should merge this or not.

matthiasbeyer:
Updated to master, will merge if CI succeeds

Issue #148, opened 2020-08-22
Asynchronous config source


szarykott:
I was wondering if there is a broader need for what can be called asnychronous config source?

At my company we use dedicated configuration service that can be asked for config for given application via HTTP. I believe there are also other solutions that use network to retrieve configuration (for instance a database source). 
While implementing `Source` trait for our configuration service I had to use blocking HTTP client. One may say this is not a problem since configuration refresh is not a frequent action. However, when you try to squeeze every ounce of performance from your application or you are using single-thread scheduler this can hurt. Also, in our specific case config is expected to be read by applications in real-time almost, so refresh should be quite frequent.

While implementing blocking HTTP source I encountered a problem with blocking `Client` from `reqwest` crate, since blocking client tries to start a Tokio runtime underneath and use non-blocking implementation, but block on it with `tokio::block_on`. Which basically panics, since if you are using tokio runtime, you are not allowed to set up another runtime inside. There are workarounds for it like:
* push config download out of runtime (but you can have hard time refreshing it then)
* use another library to perform blocking call (which can hurt on single-thread sheduler and makes this crate less friendly to use)

For those reasons I though about asynchronous configuration source.

I am afraid however that it would be hard to make it both easy to use, performant and backwards-compatible. I would drop the latter if I were to choose.

Since I come from .NET world I am quite used to `IConfiguration` introduced by .NET Core. This crate resembles it to some extent.
The idea in refactoring to use asynchoronous source could be (bit similar to `IConfiguration` having its dedicated builder):
* extend `Source` trait with `collect_async` method
* give it a default implementation using `futures::future::ready` with result of `collect` method for ease of use in case of synchronous sources (if possible)
* add `ConfigBuilder` that would **not** refresh config after every merge (might be quite ineffcient to refresh config from HTTP many times in short period of time), instead if would produce config with:
    * `build` method in case you only add synchronous sources (that returns Err if there is even single asynchronous source)
    * `build_async` method in case you have even single asynchronous source (which can build synchronous sources as well)
* library should be kept executor agnostic

What is your take on that? And, **is this crate maintained**? It looks really nice, but last comitts are quite old. I would be happy to make a pull request with aforementioned changes, but I wouldn't like to do work that would hang in a PR for eternity.

szarykott:
@matthiasbeyer since you took ownership of the crate you might be interested in one thing.

[...]

PR #180, opened 2021-03-19
ConfigBuilder (was: Add Config::with_sources())

matthiasbeyer:
This patch adds Config::with_sources() for builder-pattern multi-merge
without calling Config::refresh().
The documentation added in the patch serves as rationale.

---

I'd love to see discussion on this.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -337,7 +337,7 @@ impl ConfigBuilder {
                 ref mut sources, ..
             } => {
                 sources.append(&mut self.sources);
-            },
+            }
 
             ConfigKind::Frozen => return Err(ConfigError::Frozen),
         }
--- a/src/config.rs
+++ b/src/config.rs
@@ -100,6 +100,34 @@ impl Config {
         Ok(self)
     }
 
+    /// Merge multiple sources without refresh()ing each time.
+    ///
+    /// This function can be used if the `Config::refresh()` call would be expensive (e.g. with a
+    /// very large configuration).
+    /// Because `Config::merge()` and `Config::with_merged()` call `Config::refresh()` before
+    /// returning, and `Config::refresh()` might be expensive, this function exists.
+    ///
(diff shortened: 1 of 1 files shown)


szarykott:
As discussed in #148 this kind of code could have the desired effect. It would be easy to extend with asynchronous configuration sources if they are ever implemented just by adding `merge_async` or similar methods. 

I am thinking, however, that it might be cleaner to implement a separate struct called, for instance, `ConfigBuilder` that would produce a frozen config at the end that would be an entry point by itself, without a need to go through `Config`. Having such a separate struct would indicate clearly that workflow is different. On the other hand discoverability of such a feature would be smaller than in the case of a single entry point as it is a case with a single `Config` struct with methods on it.

Just a thought of mine on this matter.

matthiasbeyer:
So what'd you think? Merge? :smile:

matthiasbeyer:
I will merge after Fri 2nd April 2021 12:00:00 UTC if there are no complains.

matthiasbeyer:
So this is a bit off-topic for this branch, but anyways: The latest commit (which really should be squashed if we merge this) implements a builder-pattern entry point as you @szarykott suggested.

I kept the entry point for the builder on the `Config` type with `Config::builder()`, because that keeps the visibility of the whole thing. I did not (yet) remove the functions on the `Config` type itself, as I think we should rather deprecate them in one release and remove them in the following one. Changing all interface in one release seems to be a bit too big of a change for our users, IMO.

What do you think?

szarykott:
Hey, it seems ok, but I'd consider moving ConfigBuilder to a separate file as it seems a functionality of its own.

As you might have seen in #196 I'd like to separate Config and ConfigBuilder, which basically means having two structs - one with `cache`, read-only, and the other one having all mutable stuff inside. Right now they are kinda mixed as Builder just spawns a mutable Config underneath and uses its mutable methods.

Besides it might be better to merge #190 before considering merging this (due to ConfigKind::Frozen being removed).

Issue #136, opened 2020-04-18
log::Level not deserializable in 0.10.0


Yatekii:
I use [log::Level](https://docs.rs/log/0.4.10/log/enum.Level.html) in my config structs and this works great with config 0.9.0 and the serde feature enabled on the log crate (v 0.4.0).

In version 0.10.0 I get enum `Level does not have variant constructor Error', src/config/mod.rs:7:37` for every variant of the `Level` enum I try.

saethlin:
`ERROR` will work.

This is a pretty serious issue for me too, there was a _lot_ of breakage in 0.10 around case-sensitivity.

matthiasbeyer:
Do I understand correctly that this is only due to case sensitivity, but there is no real bug here?
If that's the case, I'd close this issue.

Yatekii:
I consinder this a bug tbh.

a) Why would it be case sensitive?
b) Why make this change? Enum variants are CapsCamelCase and not SCREAM_CASE.

But ofc. It is your choice :)

matthiasbeyer:
Hm. I'm not sure that's a bug in config-rs, is it? :confused:

Yatekii:
Define "bug". It is something in config-rs. https://docs.rs/log/0.4.14/src/log/lib.rs.html#513 log itself can parse it from any case.

matthiasbeyer:
Ok cool, I'll have a look!

matthiasbeyer:
I am closing this, after #188 was merged and I _think_ this issue is resolved. Feel free to comment and I'll reopen if necessary.

PR #188, opened 2021-03-27
Add test for log::Level deserialization

matthiasbeyer:
This is an attempt to tackle #136

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -53,3 +53,4 @@ glob = "0.3"
 lazy_static = "1"
 notify = "^5.0.0"
 temp-env = "0.3.0"
+log = { version = "0.4", features = ["serde"] }
--- a/tests/log.rs
+++ b/tests/log.rs
@@ -3,6 +3,7 @@ extern crate log;
 
 #[macro_use]
 extern crate serde_derive;
+extern crate serde_json;
 
 use config::*;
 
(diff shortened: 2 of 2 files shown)


TheNeikos:
I believe to make this complete case-insensitive you would need to fix the check here: https://github.com/mehcode/config-rs/blob/bf09a730d44cb72f483dab04163dd75f2b9d1746/src/de.rs#L249

As this is checked for before even the actual `log` crate gets to deserialize, thus making it case-sensitive even if the log crate might not be.

matthiasbeyer:
Okay, so I have a patch, but it does not work. It is [here](https://github.com/matthiasbeyer/config-rs/commit/33b8cf0eefffbe5ad656de1b691a668c942357f5)

It is _way_ harder to implement this, because the relevant code is deeply tied into serde and we cannot simply add settings to it. The problematic line is https://github.com/matthiasbeyer/config-rs/commit/33b8cf0eefffbe5ad656de1b691a668c942357f5#diff-be1c878c753064b7eaddf4a10ba4215a93dd1817d31e6788691cdb29752fd612R545 . I could, in theory, remove the `From` implementation and add some more powerful abstraction here, but the actual problem is not only this line, but also the `Deserialize` impl for `Value`, where we need to implement deserialization without having a `Value` object (or any other), so we cannot have settings at all.

I must say, I'm not sure whether this is possible at all, without rewriting large parts of this crate.

TheNeikos:
My question is why not let the enum try an deserialize itself? Why is the crate checking for a specific member at all? You could try constructing a StrDeserializer directly.

matthiasbeyer:
TBH I don't know yet why this is done. I hope I get the chance to investigate this.

PR #189, opened 2021-03-28
Rename get_str() → get_string() to match returned type

### LINKED ISSUE #140: Function misnamed, should corrrelate to type
Rust has types for both `str` and `String`. There are tons of functions in *std* and other crates related to the two, `as_str()`, `as_string()`, `from_string()`, `to_string()`, `to_str()`, etc.

This function is cross wired:

https://github.com/mehcode/config-rs/blob/2c0b201055be60ec45e258adcc6c4b6e01bcd627/src/config.rs#L172

It should either be named `get_string()` or return a `str` but not switch them out! Perhaps there should be two functions, one for each...

alerque:
Closes #140.

I couldn't find any good reason to make this return `&str` instead, the handling everywhere else is as `String` so just renaming the function to match what it already does seemed most logical thing to do.

This will be a breaking change to the API, but the fix will be easy. All instances of `Config.get_str()` need to be changed to `Config.get_string()`. The returned value being handled in people's code is already `String`.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -211,8 +211,8 @@ impl Config {
         }
     }
 
-    pub fn get_str(&self, key: &str) -> Result<String> {
-        self.get(key).and_then(Value::into_str)
+    pub fn get_string(&self, key: &str) -> Result<String> {
+        self.get(key).and_then(Value::into_string)
     }
 
     pub fn get_int(&self, key: &str) -> Result<i64> {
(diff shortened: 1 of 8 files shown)


matthiasbeyer:
Nice, thanks!

I'd take this patch as is, but please remove the CHANGELOG entry from the commit. I will take care of writing the changelog myself right before the release - that's just the way I prefer to do it.
Otherwise, this looks fine! :tada:

matthiasbeyer (approved):
Awesome, thanks!

PR #190, opened 2021-03-31
Remove dead code

matthiasbeyer:
This actually is twofold: First of all, it adds a `Config::freeze` function, which makes use of the `ConfigKind::Frozen` constructor, which wasn't used before.

I actually am not sure whether this is wanted, but from what I see, this is the only way that enum variant could be used, right? And because it wasn't exported, this was actually never used, so the whole "error if config is frozen" was never required in the first place.
Another option would be to remove the error handling, but I suppose this (freezing the config object) is there for a reason.

---

The second patch removes unused code and the `allow(dead_code)` lint.

---

Review welcome.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -12,36 +12,14 @@ use source::Source;
 use path;
 use value::{Table, Value, ValueKind};
 
-#[derive(Clone, Debug)]
-enum ConfigKind {
-    // A mutable configuration. This is the default.
-    Mutable {
-        defaults: HashMap<path::Expression, Value>,
-        overrides: HashMap<path::Expression, Value>,
-        sources: Vec<Box<dyn Source + Send + Sync>>,
-    },
-
-    // A frozen configuration.
-    // Configuration can no longer be mutated.
-    Frozen,
-}
-
-impl Default for ConfigKind {
-    fn default() -> Self {
-        ConfigKind::Mutable {
-            defaults: HashMap::new(),
-            overrides: HashMap::new(),
-            sources: Vec::new(),
-        }
-    }
-}
-
 /// A prioritized configuration repository. It maintains a set of
 /// configuration sources, fetches values to populate those, and provides
 /// them according to the source's priority.
 #[derive(Clone, Debug)]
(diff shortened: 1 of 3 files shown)


matthiasbeyer:
I will merge after Fri 2nd April 2021 12:00:00 UTC if there are no complains.

src/config.rs
```diff
@@ -57,6 +57,10 @@ impl Default for Config {
 }
 
 impl Config {
+    pub fn freeze(&mut self) {
```
szarykott, line 60: I'd vote against keeping this Config state.

I have serious reasons to justify it and I do have a strong vision of how config should work. After all, I spent months designing my own library. I'd like to have some opinion on that though.

In my mind there should be two separate structs:
* one that represents the builder
* one that represents built config

The reason for it is simple - moving runtime errors that can happen with Frozen / Mutable state to compilation time. Simply speaking - make invalid state impossible thanks to the type system. It has a clear advantage of "it compiles, it most probably works", but at the same time it forbids adding new sources during program operations. However, if one keeps a reference to a builder somewhere it is always possible to reuse it. From my experience, it is not a problem to do so.

src/config.rs
```diff
@@ -50,7 +28,9 @@ pub struct Config {
 impl Default for Config {
     fn default() -> Self {
         Config {
-            kind: ConfigKind::default(),
+            defaults: Default::default(),
+            overrides: Default::default(),
+            sources: Default::default(),
             cache: Value::new(None, Table::new()),
```
szarykott, line 34: Just for consistency - is there a `Default` implementation of `Value`?

szarykott (approved):

src/config.rs
matthiasbeyer: Yes, but it actually does something different than this, IIRC, because the `Default` impl for `ValueKind` is actually `ValueKind::Nil` instead of an empty table.

matthiasbeyer:
So it seems this can be merged, right? @szarykott

szarykott:
Yes, merge :)

PR #191, opened 2021-03-31
Remove unused imports

matthiasbeyer:
This just removes the unused imports as well as the lint for it.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -1,14 +1,14 @@
-use serde::de::Deserialize;
-use serde::ser::Serialize;
 use std::collections::HashMap;
 use std::fmt::Debug;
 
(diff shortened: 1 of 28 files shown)


szarykott (approved):
Looks good to me!

Tiny question - is it possible to easily squash commits before merge to keep the history clean?

matthiasbeyer:
It is and I'm going to do it right now. It's not possible with the github UI, because github sucks and does want to force its own way onto contributors, but I oppose to that strongly and try to use git as it is intended to. So I'm going to force push a cleaned up version in a minute!

matthiasbeyer:
I'm ready for merge here, if you want to merge, feel free! :smile:

PR #193, opened 2021-03-31
Misc changes

matthiasbeyer:
These are misc changes that I'd like to merge. Each patch is one change that could be debated over.

---

I'm not gonna merge this before some discussion happened.

### DIFF
--- a/src/value.rs
+++ b/src/value.rs
@@ -72,13 +72,8 @@ where
     T: Into<Value>,
 {
     fn from(values: HashMap<String, T>) -> Self {
-        let mut r = HashMap::new();
-
-        for (k, v) in values {
-            r.insert(k.clone(), v.into());
-        }
-
-        ValueKind::Table(r)
+        let t = values.into_iter().map(|(k, v)| (k, v.into())).collect();
+        ValueKind::Table(t)
     }
 }
 
--- a/src/value.rs
+++ b/src/value.rs
@@ -94,10 +94,15 @@ impl Display for ValueKind {
             ValueKind::Integer(value) => write!(f, "{}", value),
             ValueKind::Float(value) => write!(f, "{}", value),
             ValueKind::Nil => write!(f, "nil"),
-
-            // TODO: Figure out a nice Display for these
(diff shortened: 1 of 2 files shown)


src/value.rs
```diff
@@ -72,13 +72,7 @@ where
...
 {
     fn from(values: HashMap<String, T>) -> Self {
-        let mut r = HashMap::new();
-
-        for (k, v) in values {
-            r.insert(k.clone(), v.into());
-        }
-
-        ValueKind::Table(r)
+        ValueKind::Table(values.into_iter().map(|(k, v)| (k, v.into())).collect())
```
szarykott, line 75: The new version is surely much more concise! 

But on the other hand, it is harder to understand (at least for me). Maybe for the sake of readability, it could be made into two-liner like

```rust
let map = HashMap::from_iter(...);
ValueKind::Table(map)
```

And to be sure that strings are reused [`drain`](https://doc.rust-lang.org/std/collections/hash_map/struct.HashMap.html#method.drain) could be used since function takes the ownership of `values` either way? Possibly this is what `into_iter` does, but I am not sure.

szarykott (commented):
Except for the comments I left, everything is super fine!

src/value.rs
matthiasbeyer: Thanks for that comment. I will push a fixup commit for splitting the line into two shortly.

Using `HashMap::drain` is not what we want here, because it keeps the hashmap allocated for later use. `HashMap::into_iter()` takes ownership and with `collect()`ing it into a new `HashMap` object of `HashMap<String, Value>`. That's exactly what we want here, so that part is fine.

Thanks again for your comment!

src/value.rs
```diff
@@ -107,8 +96,15 @@ impl Display for ValueKind {
             ValueKind::Nil => write!(f, "nil"),
 
             // TODO: Figure out a nice Display for these
```
szarykott, line 98: Is this TODO still relevant?

szarykott (approved):

matthiasbeyer:
Because you already approved, I'm gonna squash now.

szarykott:
Feel free to merge.

PR #196, opened 2021-04-02
Create the ConfigBuilder

szarykott:
This is a draft PR to show an idea of how could ConfigBuilder look like.

It does not bring a lot of new functionality. It just splits Config that was a mutable configuration combined with a builder into two separate structs. It brings certain benefits:
* Config is no longer refreshed when a new source is added
* More from the clean code perspective, it lifts concepts of "mutable" and "frozen" config to type level, thus mitigating runtime errors. It is a clean code issue, as Config could not be frozen before as it was a dead private API.

Immutability is considered good programming practice, but not everyone might like it. I'd like to hear the opinions of users of the library - do you mutate config often after application initial bootstrap, do you consider such a mutability a needed feature?

I do not intend to push it in its current form, it is just to showcase an idea. To consider it mergeable discussion has to take place regarding few subjects:
* is it the correct way to go
* should old ways be deprecated as it is done now or should they be removed

In the current state there is a lot of deprecation warnings, so please do not mind them now and focus on more high-level aspects of this PR.

I attach a graph of the intended workflow so that it is easier to see the big picture.
image

### DIFF
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -1,0 +1,164 @@
+use std::str::FromStr;
+use std::{collections::HashMap, iter::IntoIterator};
+
+use crate::{
+    config::Config, error, error::ConfigError, path::Expression, source::Source, value::Value,
+};
+
+/// A configuration builder
+///
+/// It registers ordered sources of configuration to later build consistent [`Config`] from them.
+/// Configuration sources it defines are defaults, [`Source`]s and overrides.
+///
+/// Defaults are alaways loaded first and can be overwritten by any of two other sources.
+/// Overrides are always loaded last, thus cannot be overridden.
+/// Both can be only set explicitly key by key in code
+/// using [`set_default`](Self::set_default) or [`set_override`](Self::set_override).
+///
+/// An intermediate category, [`Source`], set groups of keys at once implicitly using data coming from external sources
+/// like files, environment variables or others that one implements. Defining a [`Source`] is as simple as implementing
+/// a trait for a struct.
+///
+/// Adding sources, setting defaults and overrides does not invoke any I/O nor builds a config.
+/// It happens on demand when [`build`](Self::build) (or its alternative) is called.
+/// Therefore all errors, related to any of the [`Source`] will only show up then.
+///
+/// # Examples
+///
+/// ```rust
+/// # use config::*;
+/// # use std::error::Error;
+/// # fn main() -> Result<(), Box<dyn Error>> {
+/// let mut builder = ConfigBuilder::default();
+///
+/// builder.set_default("default", "1")?;
+/// builder.add_source(File::new("config/settings", FileFormat::Json));
+/// builder.set_override("override", "1")?;
+///
+/// match builder.build() {
+///     Ok(config) => {
+///         // use your config
+///     },
+///     Err(e) => {
+///         // something went wrong
+///     }
+/// }
+/// # Ok(())
+/// # }
+/// ```
+///
+/// Calls can be chained as well
+/// ```rust
+/// # use std::error::Error;
+/// # use config::*;
+/// # fn main() -> Result<(), Box<dyn Error>> {
+/// let mut builder = ConfigBuilder::default();
+///
+/// builder
+///     .set_default("default", "1")?
+///     .add_source(File::new("config/settings", FileFormat::Json))
+///     .add_source(File::new("config/settings.prod", FileFormat::Json))
+///     .set_override("override", "1")?;
+/// # Ok(())
+/// # }
+/// ```
+#[derive(Debug, Clone, Default)]
+pub struct ConfigBuilder {
+    defaults: HashMap<Expression, Value>,
+    overrides: HashMap<Expression, Value>,
+    sources: Vec<Box<dyn Source + Send + Sync>>,
+}
+
+impl ConfigBuilder {
+    /// Set a default `value` at `key`
+    ///
+    /// This value can be overwritten by any [`Source`] or override.
+    ///
+    /// # Errors
+    ///
+    /// Fails if `Expression::from_str(key)` fails.
+    pub fn set_default<S, T>(&mut self, key: S, value: T) -> error::Result<&mut ConfigBuilder>
+    where
+        S: AsRef<str>,
+        T: Into<Value>,
+    {
+        self.defaults
+            .insert(Expression::from_str(key.as_ref())?, value.into());
+        Ok(self)
+    }
+
+    /// Registers new [`Source`] in this builder.
+    ///
+    /// Calling this method does not invoke any I/O. [`Source`] is only saved in internal register for later use.
+    pub fn add_source<T>(&mut self, source: T) -> &mut Self
+    where
+        T: Source + Send + Sync + 'static,
+    {
+        self.sources.push(Box::new(source));
+        self
+    }
+
+    /// Set an override
+    ///
+    /// This function sets an overwrite value. It will not be altered by any default or [`Source`]
+    ///
+    /// # Errors
+    ///
+    /// Fails if `Expression::from_str(key)` fails.
+    pub fn set_override<S, T>(&mut self, key: S, value: T) -> error::Result<&mut ConfigBuilder>
+    where
+        S: AsRef<str>,
+        T: Into<Value>,
+    {
+        self.overrides
+            .insert(Expression::from_str(key.as_ref())?, value.into());
+        Ok(self)
+    }
+
+    /// Reads all registered [`Source`]s.
+    ///
+    /// This is the method that invokes all I/O operations.
+    /// For a non consuming alternative see [`build_cloned`](Self::build_cloned)
+    ///
+    /// # Errors
+    /// If source collection fails, be it technical reasons or related to inability to read data as `Config` for different reasons,
+    /// this method returns error.
+    pub fn build(self) -> error::Result<Config> {
+        Self::build_internal(self.defaults, self.overrides, &self.sources)
+    }
+
+    /// Reads all registered [`Source`]s.
+    ///
+    /// Similar to [`build`](Self::build), but it does not take ownership of `ConfigBuilder` to allow later reuse.
+    /// Internally it clones data to achieve it.
+    ///
+    /// # Errors
+    /// If source collection fails, be it technical reasons or related to inability to read data as `Config` for different reasons,
+    /// this method returns error.
+    pub fn build_cloned(&self) -> error::Result<Config> {
+        Self::build_internal(self.defaults.clone(), self.overrides.clone(), &self.sources)
+    }
+
+    fn build_internal(
+        defaults: HashMap<Expression, Value>,
+        overrides: HashMap<Expression, Value>,
+        sources: &Vec<Box<dyn Source + Send + Sync>>,
+    ) -> error::Result<Config> {
+        let mut cache: Value = HashMap::<String, Value>::new().into();
+
+        // Add defaults
+        for (key, val) in defaults.into_iter() {
+            key.set(&mut cache, val);
+        }
+
+        // Add sources
+        sources.collect_to(&mut cache)?;
+
+        // Add overrides
+        for (key, val) in overrides.into_iter() {
+            key.set(&mut cache, val);
+        }
+
+        Ok(Config::new(cache))
+    }
+}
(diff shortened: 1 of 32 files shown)


src/config.rs
```diff
@@ -57,45 +58,32 @@ impl Default for Config {
...
-            } => {
-                sources.push(Box::new(source));
-            }
-
-            ConfigKind::Frozen => {
-                return Err(ConfigError::Frozen);
-            }
-        }
-
+        self.push(source)?;
```
matthiasbeyer, line 75: Why are you changing this implementation? There's no need to in this PR, AFAICS.

src/config.rs
```diff
@@ -57,45 +58,32 @@ impl Default for Config {
...
```
matthiasbeyer, line 86: Why are you changing this implementation? There's no need to in this PR, AFAICS.

src/config.rs
```diff
@@ -139,7 +128,29 @@ impl Config {
...
+            } => {
+                sources.push(Box::new(source));
+                return Ok(())
+            }
+
+            ConfigKind::Frozen => {
+                return Err(ConfigError::Frozen);
+            }
+        }
+    }
```
matthiasbeyer, line 150: Please also remove this function when cleaning up my above comments.

src/builder.rs
```diff
@@ -0,0 +1,149 @@
...
+
+match builder.build() {
+    Ok(config) => {
+        // use your config
+    },
+    Err(e) => {
+        // something went wrong
+    }
+}
+```
```
matthiasbeyer, line 46: Two things: I like that the builder takes `&mut self`, this way it can be chained and called like you did in the code example above. I'd like to ask, though, whether you could add another example where you chain the calls.

Also, I think it would be a good idea to provide a `Config::builder()` function, as layed out in #180 for example. Good for discoverability and not too much overhead on our side.

src/builder.rs
```diff
@@ -0,0 +1,149 @@
+use crate::{
+    config::Config,
+    value::Value,
+    source::Source,
+    error,
+    error::ConfigError,
+    path::Expression
+};
+use std::collections::HashMap;
```
matthiasbeyer, line 9: Please import from `std` first (you'll need a blank line between the two import blocks).

matthiasbeyer (changes requested):

src/builder.rs
```diff
@@ -0,0 +1,149 @@
...
+
+/**
```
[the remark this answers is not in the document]
szarykott: If this is not a matter of taste, then according to [Rust Reference](https://doc.rust-lang.org/reference/comments.html) doc comments are line comments that start with `///` or block comments that use `/** ... */` syntax (double asterisk at the beginning).

> Line doc comments beginning with exactly three slashes (///), and block doc comments (/** ... */), both inner doc comments, are interpreted as a special syntax for doc attributes. 

I did not know about it until VSCode suggested this and it actually works, I tested with `cargo doc`. Thought this is convinient for multiline doc comments.

src/builder.rs
```diff
@@ -0,0 +1,149 @@
...
+    this method returns error.
+    
+    It does not take ownership of `ConfigBuilder` to allow later reuse. Internally it clones data to achieve it.
+    */
+    pub fn build(&self) -> error::Result<Config>  {
+        let mut cache: Value = HashMap::<String, Value>::new().into();
+
+        // Add defaults
+        for (key, val) in self.defaults.iter() {
+            key.set(&mut cache, val.clone());
```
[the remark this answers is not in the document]
szarykott: Point is that `build` takes `&self`, therefore I cannot use methods that consume struct fields.

It was a conscious decision to do so - as it is written in the comment. I can provide an additional method called for instance `build_owned` that would consume `self`. It would give users the choice - if they do not care about reusing the builder they could just use it. On the other hand, if they do care about it, because they expect configuration to be refreshed, they could stick to `build`. 

Or I could change this method to `build_cloned` and name the one taking ownership `build`?

src/builder.rs
```diff
@@ -0,0 +1,149 @@
...
+        }
+
+        // Add sources
+        self.sources.collect_to(&mut cache)?;
+
+        // Add overrides
+        for (key, val) in self.overrides.iter() {
+            key.set(&mut cache, val.clone());
```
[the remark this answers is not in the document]
szarykott: I answered in the other thread.

src/builder.rs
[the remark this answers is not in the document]
matthiasbeyer: Yes, the reference is the syntax. In [the official styleguide](https://doc.rust-lang.org/1.0.0/style/style/comments.html) they use `//` instead of `/*` (andn `///` instead of `/**`).

src/builder.rs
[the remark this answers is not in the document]
matthiasbeyer: I doubt there is a use-case for not building an owned version, hence I'd use `pub fn build(self) -> ...`.

Also, I'm not sure whether it is possible to have a `fn build(self)` if the previous builder functions only return `&mut Self` instead of `Self` (see #180).

src/builder.rs
```diff
@@ -0,0 +1,149 @@
...
+    /**
+    Creates an empty `ConfigBuilder` without any configuration source registered.
+
+    This is the [`Default`] implementation of `ConfigBuilder`
+    ```rust
+    # use config::ConfigBuilder;
+    let builder = ConfigBuilder::empty();
+    ```
+    */
+    pub fn empty() -> Self {
```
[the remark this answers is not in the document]
szarykott: I removed it and used `[derive(Default)]` instead of it on a struct.

src/builder.rs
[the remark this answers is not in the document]
szarykott: I added `build_cloned` and applied your suggestion in `build`.

src/config.rs
szarykott: Reverted to the old state.

src/config.rs
szarykott: Reverted to the old state.

src/config.rs
szarykott: Done.

src/builder.rs
szarykott: Done :)

src/builder.rs
szarykott: Done :)

src/builder.rs
matthiasbeyer: :+1: awesome. But you need to `build()` it, too! :laughing:

src/builder.rs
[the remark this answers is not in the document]
szarykott: Let it be. Changed.

src/builder.rs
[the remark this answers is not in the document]
szarykott: And regarding usefulness - for example in my company we do use a configuration like this in a dynamic way. It makes little sense when all the Sources are files and environment, but as soon as you start adding remote sources be it custom, Consult, or whatever you want, it starts making sense to refresh configuration from time to time in order to take the newest configuration into account.

matthiasbeyer:
Hi. So how do we move forward here? I guess this and #180 overlaps too much to keep both open, right? I'd close #180 therefore and continue with this PR here (not only, but also because I want to encourage you to contribute! :laughing: ).

If you don't mind, feel free to take patches from [git.beyermatthias.de/config - branch 'szarykott-builder-patches'](https://git.beyermatthi.as/config/log/?h=szarykott-builder-patches) (`git fetch https://git.beyermatthi.as/config szarykott-builder-patches && git merge --ff-only FETCH_HEAD`) were I did some minor cleanup for this PR for a few things I think are appropriate. The change to calling `Expression::from_str` is controversial, I guess :smile: but I'd rather be a bit more explicit in the code here.

szarykott:
Thank you for the heads up @matthiasbeyer !

Correct, those two PRs follow the same principle with minor differences. Thank you for willingness to merge my PR then. I'll see what are the changes you propose and push them here. I will handle it, but it might take me few days.

matthiasbeyer:
No worries, we are not in a hurry! :laughing:

src/builder.rs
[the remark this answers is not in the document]
szarykott: Are you okay with the current state?

szarykott:
I merged your patches and pushed them. Could you please resolve all pull request discussions you consider solved for better visibility of what is there to do?

If you are ok with the way it is, I am also going to refactor tests to use builder pattern, due to a large number of deprecation warnings in tests.

szarykott:
I tried to squash commits, something weird happened, please be patient ;)

matthiasbeyer:
I resolved everything and I will do a new review once you're ready! :smile:

szarykott:
Hello again!

**I think this is more-less the final version of this pull request.**

There are a lot of files changed, but most of them are tests.

Here's a brief description of the changes.
1. builder.rs contains a builder struct that is the essence of this pull request.
2. Old implementation is deprecated, but not removed.
3. To avoid deprecation warnings, I added allow(deprecated) on appropriate lines in the crate.
4. To have new features tested I changed all the tests to use ConfigBuilder.
5. To have old features still tested I duplicated appropriate tests and put in tests/legacy folder (**hoping they will be removed after 0.12 release**). Tests in this folder have `allow(deprecated)` attribute. The entry point for them is the legacy_tests.rs file in the tests folder.
6. I added a new implementation of Source (for [Box<...>]) as [Clippy issued a warning regarding the `build_internal` method](https://rust-lang.github.io/rust-clippy/master/index.html#ptr_arg). Another way to fix it would be to disable this lint on this method, but it seems a bit hacky, while it was relatively easy to add a new implementation of Source.

src/builder.rs
```diff
@@ -0,0 +1,162 @@
...
+
+impl ConfigBuilder {
+    /// Set a default `value` at `key`
+    ///
+    /// This value can be overwritten by any [`Source`] or override.
+    ///
+    /// # Errors
+    ///
+    /// Fails if `Expression::from_str(key)` fails.
+    pub fn set_default<S, T>(&mut self, key: S, value: T) -> error::Result<&mut ConfigBuilder>
```
matthiasbeyer, line 78: I'm thinking about why we're using the `&mut self` builder-pattern style here. Maybe `mut self` would be more appropriate (moving and returning `Self`), so we can chain the `build()` function properly?

Nothing that blocks this PR from landing though!

That would also make the `impl Source for [Box<dyn Source + Send + Sync>]` unnecessary I guess.

matthiasbeyer (approved):
Awesome.

One comment, but that wouldn't block merging. I rather merge now and do some more changes later :smiley: 

One more rebase please:

1. reword the first commit to just "Add ConfigBuilder for builder-pattern style Config initialization"
2. squash into first commit: b2e50ff
3. squash 43973fe and fc0b227 into 6d2dada, rewording to "Use builder-pattern in tests, keep old interface tests in legacy" (or something like that)
4. drop de097d3

You can keep the other commits, especially 2f8aca7 (but, if possible, with some more appropriate wording) - because it can easily be reverted later.

Sorry for another round of rebasing, I hope you don't mind. Otherwise, this is ready for merging IMO.

src/builder.rs
szarykott: Actually, I've used `&mut self`, because this is how it was used in `Config`.

But point is valid, consuming version would allow to chain things properly, I am going to change this PR. After all if it turns out in the future that `&mut self` version is required by users, it is "less breaking" to lessen required ownership (self -> &self, etc.) than the other way around.

It would not make this implementation you mention unnecessary as it is needed because there is `build_cloned` function that takes `&self`. It makes sense to have a cloned version especially when speaking about configuration that can change over time (etcd, consul, even files in rare circumstances) that you might want refresh periodically in your application.

matthiasbeyer: > files in rare circumstances

not even in rare ones: Think of a server software which should be able to pick up config changes without restarting!

> I am going to change this PR

Awesome!

szarykott: I am happy you agree about the two flavors of the build method. I did not want to sound too pushy on this as I remember your opposition to the idea in some earlier comments, but maybe I just recall wrongly. 

Code is changed, ConfigBuilder's methods are not consuming, except for one of two build methods. Feel free to merge as soon as you want!

szarykott: I did not change commit messages exactly as you wanted, but I hope it is fine the way it is.

szarykott:
@matthiasbeyer merge when you want. I rebased my code on master and added one commit which modifies env.rs file in similar way that previous code did to other tests files (use builder in main tests and keep old tests in legacy_tests).

matthiasbeyer:
Awesome. I really love the progress in this crate lately and the many people contributing! Thank you!

PR #202, opened 2021-04-19
Add support for RON format

### LINKED ISSUE #116: [Feature request] Ron support
Would be nice if this crate supported [Rusty Object Notation](https://github.com/ron-rs/ron)

skreborn:
This change set adds support for the [RON](https://github.com/ron-rs/ron) format.

The files are mostly copied and modified slightly from the existing format parsers, with the addition of handling the unit, character, and optional types.

Resolves #116.

### DIFF
--- a/src/file/format/ron.rs
+++ b/src/file/format/ron.rs
@@ -1,0 +1,68 @@
+use std::collections::HashMap;
+use std::error::Error;
+
+use ron;
+
+use crate::value::{Value, ValueKind};
+
+pub fn parse(
+    uri: Option<&String>,
+    text: &str,
+) -> Result<HashMap<String, Value>, Box<dyn Error + Send + Sync>> {
+    let value = from_ron_value(uri, ron::from_str(text)?)?;
+    match value.kind {
+        ValueKind::Table(map) => Ok(map),
+
+        _ => Ok(HashMap::new()),
+    }
+}
+
+fn from_ron_value(
+    uri: Option<&String>,
+    value: ron::Value,
+) -> Result<Value, Box<dyn Error + Send + Sync>> {
+    let kind = match value {
+        ron::Value::Option(value) => match value {
+            Some(value) => from_ron_value(uri, *value)?.kind,
+            None => ValueKind::Nil,
+        },
+
+        ron::Value::Unit => ValueKind::Nil,
+
+        ron::Value::Bool(value) => ValueKind::Boolean(value),
+
+        ron::Value::Number(value) => match value {
+            ron::Number::Float(value) => ValueKind::Float(value.get()),
+            ron::Number::Integer(value) => ValueKind::Integer(value),
+        },
+
+        ron::Value::Char(value) => ValueKind::String(value.to_string()),
+
+        ron::Value::String(value) => ValueKind::String(value),
+
+        ron::Value::Seq(values) => {
+            let array = values
+                .into_iter()
+                .map(|value| from_ron_value(uri, value))
+                .collect::<Result<Vec<_>, _>>()?;
+
+            ValueKind::Array(array)
+        }
+
+        ron::Value::Map(values) => {
+            let map = values
+                .iter()
+                .map(|(key, value)| -> Result<_, Box<dyn Error + Send + Sync>> {
+                    let key = key.clone().into_rust::<String>()?;
+                    let value = from_ron_value(uri, value.clone())?;
+
+                    Ok((key, value))
+                })
+                .collect::<Result<HashMap<_, _>, _>>()?;
+
+            ValueKind::Table(map)
+        }
+    };
+
+    Ok(Value::new(uri, kind))
+}
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,7 +14,7 @@ license = "MIT/Apache-2.0"
 maintenance = { status = "actively-developed" }
 
 [features]
-default = ["toml", "json", "yaml", "hjson", "ini"]
+default = ["toml", "json", "yaml", "hjson", "ini", "ron"]
 json = ["serde_json"]
 yaml = ["yaml-rust"]
 hjson = ["serde-hjson"]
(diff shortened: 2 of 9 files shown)


src/file/format/ron.rs
```diff
@@ -0,0 +1,74 @@
...
+
+        _ => Ok(HashMap::new()),
+    }
+}
+
+fn from_ron_value(
+    uri: Option<&String>,
+    value: &ron::Value,
+) -> Result<Value, Box<dyn Error + Send + Sync>> {
+    Ok(match *value {
```
matthiasbeyer, line 24: Can you please restructure this function implementation to be a bit more one step at a time? :laughing: 

* First of all, you could `use ron::Value` to eliminate the full qualification path (`run::Value::Bool` -> `Value::Bool`).
* Please do use variables (not `Ok(everything)`, but `let v = match { /*...*/ }; Ok(v)`.

src/file/format/ron.rs
```diff
@@ -0,0 +1,74 @@
...
+    match value.kind {
+        ValueKind::Table(map) => Ok(map),
+
+        _ => Ok(HashMap::new()),
+    }
+}
+
+fn from_ron_value(
+    uri: Option<&String>,
+    value: &ron::Value,
```
matthiasbeyer, line 22: Why is `value` passed by reference? From the calling site it looks like we can move it into the function and save a few `clone()` calls here and there, can't we?

src/file/format/ron.rs
matthiasbeyer: Ah, you could even let the `match` return the `kind` only and have `Ok(Value::new(uri, kind))` at the end of the function!

src/file/format/ron.rs
```diff
@@ -0,0 +1,74 @@
...
+        ron::Value::Char(ref value) => Value::new(uri, ValueKind::String(value.to_string())),
+
+        ron::Value::String(ref value) => Value::new(uri, ValueKind::String(value.clone())),
+
+        ron::Value::Seq(ref values) => {
+            let mut array = Vec::new();
+
+            for value in values.iter() {
+                array.push(from_ron_value(uri, value)?);
+            }
```
matthiasbeyer, line 56: Please use an iterator adaptor: `values.iter().map(|v| from_ron_value(uri, v)).collect::<Result<Vec<_>>()?`. I like to believe that it performs better because it can preallocate the `Vec<_>` with the appropriate size.

(or with `into_iter()` if you change the function interface of `from_ron_value()`, of course).

src/file/format/ron.rs
```diff
@@ -0,0 +1,74 @@
...
+
+        ron::Value::Map(ref values) => {
+            let mut map = HashMap::new();
+
+            for (key, value) in values.iter() {
+                map.insert(
+                    key.clone().into_rust::<String>()?,
+                    from_ron_value(uri, value)?,
+                );
+            }
```
matthiasbeyer, line 69: Same as above applies here.

matthiasbeyer (changes requested):
Hi!

Thank you very much for your PR, I love that people contribute to this crate!

I found a few things we have to discuss, feel free to ask questions or teach me where I'm wrong! :laughing:

src/file/format/ron.rs
skreborn: I've modified the `match` to return the correct `Kind` as asked.

For the qualification, I'd rather keep the `ron::` prefix as to differentiate it from `config`'s own `Value` type.

src/file/format/ron.rs
skreborn: It was passed by reference only because it was like that in the JSON format. I've modified it to take an owned value.

src/file/format/ron.rs
skreborn: Done!

src/file/format/ron.rs
skreborn: Done, but the cloning unfortunately stayed because `ron::Map` does not implement `into_iter`, so I couldn't do much about that.

skreborn:
I've addressed your comments, hopefully the changes are better now. Let me know if there's anything else!

src/file/format/ron.rs
```diff
@@ -19,56 +19,51 @@ pub fn parse(
...
+        ron::Value::Map(values) => {
+            let map = values
+                .iter()
+                .map(|(key, value)| match key.clone().into_rust::<String>() {
+                    Ok(key) => match from_ron_value(uri, value.clone()) {
+                        Ok(value) => Ok((key, value)),
+                        Err(err) => Err(err),
+                    },
+                    Err(err) => Err(err.into()),
+                })
```
matthiasbeyer, line 61: How about

```rust
.into_iter() // use into_iter() here so you don't have to clone()
.map(|(key, value)| {
    let k = key.into_rust::<String>()?;
    let v = from_ron_value(uri, value)?;
    Ok((k, v))
})
```

matthiasbeyer (changes requested):

src/file/format/ron.rs
matthiasbeyer: ah, okay. So ignore that part of my latest comment please :laughing:

src/file/format/ron.rs
matthiasbeyer: Agreed.

src/file/format/ron.rs
skreborn: Unfortunately, `ron::Map` does not implement `into_iter`, so I can't do much about the cloning. But I have no idea why I used matches when I could've just let Rust handle the early returns for me. It's way too early in the morning.

matthiasbeyer (approved):
Nice. Thanks a lot for your contribution. Please squash the commits into one and then I think we're ready to go!

skreborn:
> Please squash the commits into one and then I think we're ready to go!

Squashed and ready!

PR #207, opened 2021-06-01
Add AsyncSource with tests, docs and examples

### LINKED ISSUE #148: Asynchronous config source

szarykott:
Resolves #148 

This pull request adds a new type of source - `AsyncSource`, that supports `async fn`. Also, it modifies the builder in order to handle it properly. Builder implementation based on [this article](http://cliffle.com/blog/rust-typestate/).

There is no implementation of such a source inside the library since there are various runtimes to choose from as explained in the rustdocs in this pull request.

One minor change - the type of `Result` in regular `Source` has been changed from type synonym defined inside the library to std's `Result` since ... type synonym was not exported from the crate.

Json5 tests have been modified for Windows. There was some issue with line endings I was not able to resolve in error message tests. CI for this package runs on Linux, so test will run.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -40,3 +40,7 @@ json5_rs = { version = "0.3", optional = true, package = "json5" }
 serde_derive = "1.0.8"
 float-cmp = "0.8"
 chrono = { version = "0.4", features = ["serde"] }
+tokio = { version = "1", features = ["rt-multi-thread", "macros", "fs", "io-util", "time"]}
+warp = "0.3.1"
+futures = "0.3.15"
+reqwest = "0.11.3"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,6 +23,7 @@ ini = ["rust-ini"]
 json5 = ["json5_rs"]
 
 [dependencies]
+async-trait = "0.1.50"
 lazy_static = "1.0"
 serde = "1.0.8"
 nom = "6"
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -1,7 +1,8 @@
 use std::str::FromStr;
 use std::{collections::HashMap, iter::IntoIterator};
 
-use crate::{config::Config, error, path::Expression, source::Source, value::Value};
+use crate::error::Result;
+use crate::{config::Config, path::Expression, source::Source, value::Value};
 
 /// A configuration builder
 ///
@@ -72,7 +73,7 @@ impl ConfigBuilder {
     /// # Errors
     ///
     /// Fails if `Expression::from_str(key)` fails.
-    pub fn set_default<S, T>(mut self, key: S, value: T) -> error::Result<ConfigBuilder>
+    pub fn set_default<S, T>(mut self, key: S, value: T) -> Result<ConfigBuilder>
     where
         S: AsRef<str>,
         T: Into<Value>,
@@ -100,7 +101,7 @@ impl ConfigBuilder {
     /// # Errors
     ///
     /// Fails if `Expression::from_str(key)` fails.
-    pub fn set_override<S, T>(mut self, key: S, value: T) -> error::Result<ConfigBuilder>
+    pub fn set_override<S, T>(mut self, key: S, value: T) -> Result<ConfigBuilder>
     where
         S: AsRef<str>,
         T: Into<Value>,
@@ -118,7 +119,7 @@ impl ConfigBuilder {
     /// # Errors
     /// If source collection fails, be it technical reasons or related to inability to read data as `Config` for different reasons,
     /// this method returns error.
-    pub fn build(self) -> error::Result<Config> {
+    pub fn build(self) -> Result<Config> {
         Self::build_internal(self.defaults, self.overrides, &self.sources)
     }
 
@@ -130,7 +131,7 @@ impl ConfigBuilder {
     /// # Errors
     /// If source collection fails, be it technical reasons or related to inability to read data as `Config` for different reasons,
     /// this method returns error.
-    pub fn build_cloned(&self) -> error::Result<Config> {
+    pub fn build_cloned(&self) -> Result<Config> {
         Self::build_internal(self.defaults.clone(), self.overrides.clone(), &self.sources)
     }
 
@@ -138,7 +139,7 @@ impl ConfigBuilder {
         defaults: HashMap<Expression, Value>,
         overrides: HashMap<Expression, Value>,
         sources: &[Box<dyn Source + Send + Sync>],
-    ) -> error::Result<Config> {
+    ) -> Result<Config> {
         let mut cache: Value = HashMap::<String, Value>::new().into();
 
         // Add defaults
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -2,6 +2,7 @@ use std::str::FromStr;
 use std::{collections::HashMap, iter::IntoIterator};
 
 use crate::error::Result;
+use crate::source::AsyncSource;
 use crate::{config::Config, path::Expression, source::Source, value::Value};
 
 /// A configuration builder
@@ -22,15 +23,24 @@ use crate::{config::Config, path::Expression, source::Source, value::Value};
 /// It happens on demand when [`build`](Self::build) (or its alternative) is called.
 /// Therefore all errors, related to any of the [`Source`] will only show up then.
 ///
+/// # Sync and async builder
+///
+/// [`ConfigBuilder`] uses type parameter to keep track of builder state.
+///
+/// In [`DefaultState`] builder only supports [`Source`]s
+///
+/// In [`AsyncState`] it supports both [`Source`]s and [`AsyncSource`]s at the price of building using `async fn`.
+///
 /// # Examples
 ///
 /// ```rust
 /// # use config::*;
 /// # use std::error::Error;
 /// # fn main() -> Result<(), Box<dyn Error>> {
-/// let mut builder = ConfigBuilder::default()
+/// let mut builder = Config::builder()
 ///     .set_default("default", "1")?
 ///     .add_source(File::new("config/settings", FileFormat::Json))
+/// //  .add_async_source(...)
 ///     .set_override("override", "1")?;
 ///
 /// match builder.build() {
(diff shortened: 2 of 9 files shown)


Cargo.toml
```diff
@@ -1,18 +1,18 @@
...
 license = "MIT/Apache-2.0"
-edition = "2018"
+name = "config"
+readme = "README.md"
+repository = "https://github.com/mehcode/config-rs"
+version = "0.11.0"
 
 [badges]
-maintenance = { status = "actively-developed" }
+maintenance = {status = "actively-developed"}
```
matthiasbeyer, line 15: All of the changes above are not necessary for the PR, please revert them back!

Cargo.toml
```diff
@@ -23,9 +23,10 @@ ini = ["rust-ini"]
 json5 = ["json5_rs"]
 
 [dependencies]
+async-trait = "0.1.50"
 lazy_static = "1.0"
-serde = "1.0.8"
 nom = "6"
+serde = "1.0.8"
```
matthiasbeyer, line 29: This diff is also not necessary, please revert it (moving the `serde` line).

Cargo.toml
```diff
@@ -36,6 +37,10 @@ ron = { version = "0.6", optional = true }
 json5_rs = { version = "0.3", optional = true, package = "json5" }
 
 [dev-dependencies]
-serde_derive = "1.0.8"
+chrono = {version = "0.4", features = ["serde"]}
 float-cmp = "0.8"
-chrono = { version = "0.4", features = ["serde"] }
```
matthiasbeyer, line 41: Moving the `chrono` and `serde_derive` lines is also not necessary.

src/builder.rs
```diff
@@ -109,6 +155,44 @@ impl ConfigBuilder {
...
+    where
+        T: AsyncSource + Send + Sync + 'static,
+    {
+        let async_state = ConfigBuilder {
+            state: AsyncState {
+                sources: self
+                    .state
+                    .sources
+                    .into_iter()
+                    .map(|s| SourceType::Sync(s))
```
matthiasbeyer, line 187: Minor nit: `.map(SourceType::Sync)` should work here :laughing:

src/builder.rs
```diff
@@ -157,3 +245,85 @@ impl ConfigBuilder {
...
+        for (key, val) in defaults.into_iter() {
+            key.set(&mut cache, val);
+        }
+
+        for source in sources.iter() {
+            match source {
+                SourceType::Sync(source) => source.collect_to(&mut cache)?,
+                SourceType::Async(source) => source.collect_to(&mut cache).await?,
+            }
+        }
```
matthiasbeyer, line 320: This is nice for now, but I guess we can improve this later so that the async sources are collected in parallel and the sync sources during that!

src/source.rs
```diff
@@ -12,23 +14,65 @@ pub trait Source: Debug {
 
     /// Collect all configuration properties available from this source and return
     /// a HashMap.
-    fn collect(&self) -> Result<HashMap<String, Value>>;
+    fn collect(&self) -> Result<HashMap<String, Value>, ConfigError>;
```
matthiasbeyer, line 17: Any particular reason why you don't use the `crate::error::Result<_>` anymore? If no, please revert this (and all similar) changes, so we don't have to specify `ConfigError` all the time in this module!

tests/file_json5.rs
```diff
@@ -65,6 +66,7 @@ fn test_file() {
 }
 
 #[test]
+#[cfg(not(target_os = "windows"))] // there is something wrong with CR LF I cannot solve
```
matthiasbeyer, line 69: @up9cloud maybe you have an idea what's wrong here?

matthiasbeyer (changes requested):
Hi!

Overall this looks really nice and I like that you added a few tests as well! I hope we can get this in!

I have a few things to remark, though! :smile:

tests/file_json5.rs
up9cloud: @matthiasbeyer Well, I guess the error message from crate `pest` (parser for json5-rs) uses different EOL.

@szarykott try this?
```diff
- " --> 2:7\n  |\n2 |   ok: true␊\n  |       ^---\n  |\n  = expected null in {}"
+ " --> 2:7\n  |\n2 |   ok: true␍␊\n  |       ^---\n  |\n  = expected null in {}"
```

see https://github.com/pest-parser/pest/blob/b2c350862f52f3b51f6a32c79727e3dec3a408ad/pest/src/error.rs#L514

up9cloud: @matthiasbeyer Well, after review the pest (parser for json5-rs) code base, I guess @szarykott didn't set `core.autocrlf`  to `false`.

szarykott:
Thank you for the review! I'll try to fix things and answer in a few days.

Cargo.toml
szarykott: Done

Cargo.toml
szarykott: Done

src/builder.rs
szarykott: Indeed, this would be a nice improvement, but this would increase the complexity of the solution, so for the first version, I'd prefer to stick to the current solution.

src/builder.rs
szarykott: Changed

src/source.rs
szarykott: I saw that `Result` is not exported from the crate, which I remember vaguely did not play nicely with rust-analyzer that automatically used crate's method signature when asked to "implement trait functions". And since it wasn't exported, it didn't compile, I had to think about why. It was a long time ago, so it might be fixed now.

And personally, I am not a big fan of exporting Result, because in the case of import `use config::*` it overwrites std's `Result` and it is annoying. Hence the decision to change it here, since people have to implement it.

I am open to discussion though, it might be good idea to test it once more as well.

tests/file_json5.rs
szarykott: Code is reverted, I'll try to fit it on my side and rely on GitHub's CI/CD to make sure everything is ok.

Cargo.toml
szarykott: Done

szarykott:
It seems there is a problem - due to `const fn` usage in `reqwest` dependency `socket2` it fails on Rust 1.44.

I see few options:
* try to figure out how to compile code that requires it only on new Rust
* do not run CI on 1.44
* try to play with dependencies to lower socket2 version transitively (requires to drop to `tokio` < 1 in examples)

Feel free to input your suggestions and in the meantime, I'll try to figure it out.

matthiasbeyer:
Well, if we need a newer rust version because we need reqwest, the sensible thing would be to upgrade our MSRV to match that of reqwest.

From what I see, though, this crate is only required in the examples. I would therefore limit the requirement in the `Cargo.toml` so  that it is only pulled in for the examples but not for the actual crate. If that is possible, we could exclude the examples from the 1.44 job and only compile them with stable rust.

What do you think?

matthiasbeyer:
Try to cherry-pick https://github.com/matthiasbeyer/config-rs/commit/6f84f869e68ec87558361808d8fadbb95aecbf16 onto your branch if you like. This _should_ fix the issue. (CI run [here](https://github.com/matthiasbeyer/config-rs/actions/runs/952365993))

szarykott:
Green. I dumped versions of packages that are only used in examples and tests to older versions that do not use fancy language features.

matthiasbeyer:
Yeah, that change was unnecessary please pull (fast-forward) from https://github.com/matthiasbeyer/config-rs/tree/pr-207 (or, if you don't mind, I can directly push my changes to this PR).

I also had a quick look at the commit history and I'm not happy yet, sorry. As soon as I'm happy with the overall changes and did a review, we should clean the history. There are a lot of unnecessary changes and commits in the history which should be removed. **I can offer you help with that, even doing the rebase myself if you are okay with it / don't have the time / whatever!**

---

Over all, though, I want to thank you very much for your patience and work on this crate! Async sources really is a feature we want to have here and I'm only so nit-picky here so we are doing the first steps right so we don't have to rewrite the feature later on (and that bit about the commit history is just my OCD on git history kicking in :laughing: )! Please don't be offended by my nit-picking!

szarykott:
@matthiasbeyer I'd be glad if you fixed the history in a way you see appropriate! Same goes for pushing the code you mentioned from your branch.

matthiasbeyer:
Feel free to review what I did and check whether I messed up somewhere... now this seems to be ready for merging (I'll do another review next week... too much to do today and in the next two days).

szarykott:
It looks good to me

matthiasbeyer:
Found some more things. Somehow the old `tokio` version slipped through my rebase, which is fixed now. Also, I simplified the example (also to showcase error conversion).

szarykott:
Thank you. I pushed a commit that fixes `rust fmt` check, but I am not sure how to fix (and why it is illegal) the one that blocks fixup commits :(

matthiasbeyer:
This PR has come a long way and I really, really appreciate the work you've put in here.
Thanks a lot!

I'm merging now! :tada:

Issue #62, opened 2018-02-13
Expose ValueKind


j-t-d:
Would it be possible to expose ValueKind from value.rs to allow some introspection?

ie, change: pub use value::Value; in lib.rs to
pub use value::{Value, ValueKind};

XX:
I alse need access to ValueKind to implement serialization of my settings struct to config.

mehcode:
I'll be removing `ValueKind` in `0.10` (and replacing it with just `Value`). Switching to new method of tracking errors (which was the reason for separation).

greenwoodcm:
It looks like `ValueKind` still exists on 0.10.1.  I still need it because I want to match on the contained type and value of a `Value`.  I was going to hack around it by invoking `into_bool`, `into_int`, etc one by one, but that seems to not work because those will coerce the value if possible.  For instance, I can't distinguish between `Boolean(true)` and `Integer(1)`.

gavinleroy:
@mehcode I'm trying to implement the `Source` trait for an object and I'm unable to create a `Value` due to the privacy of `ValueKind`. 

Could we expose `ValueKind`?

matthiasbeyer:
Is this still desired in 0.11? I'm hesitant to make `ValueKind` public, actually.

polivbr:
I'm looking to write a source that sits on top of a config and does variable interpolation, and having access to ValueKind would make this much easier, as I need to recurse through the tree of values.

matthiasbeyer:
What does your approach look like? IMO you could load the config as a `struct LoadedConfig` and then do the interpolation from there `LoadedConfig::interpolate(variables)? // returns Result<Config>`, after the configuration object is loaded from `config`. For that you'd not need access to the `ValueKind`!

Or am I missing something?

polivbr:
How do I recurse across the hierarchy of values in the Config, performing interpolation on the string values, regardless of how deep they are in the hierarchy? If I had access to ValueKind, I can match on the type, processing strings, ignoring other scalar types, and recursively process arrays and maps. Not sure how I would handle this without ValueKind. 

Can I ask what the objection to exposing this is?

matthiasbeyer:
The objection is mainly as it makes interior from the implementation public. This would mean that changing the implementation of the crate is always a breaking change.

You would implement the interpolation on your config structs, so there wouldn't be any recursion involved!

polivbr:
[...]

PR #208, opened 2021-06-03
ValueKind is now public to users of the crate

mjte-riot:
`ValueKind` is now exposed publicly to users of this crate. This is not necessary for normal operations on `Value`, but can be useful for operations such as serializing configuration to another format, such as JSON.

This is related to issue #62


matthiasbeyer (approved):

matthiasbeyer:
I think this is appropriate. Lets see what the future brings with this exposed.

PR #209, opened 2021-06-08
refactor(env): optimize and reduce allocations

eduardocanellas:
I had a slow test involving config and I thought that optimizing this module would make it better.

Unfortunately for me, the test is still slow after these changes (the problem is probably elsewhere instead of config), but I thought it was worth opening a PR anyway, as you may want to add them.

### DIFF
--- a/src/env.rs
+++ b/src/env.rs
@@ -83,21 +83,14 @@ impl Source for Environment {
         let mut m = HashMap::new();
         let uri: String = "the environment".into();
 
-        let separator = match self.separator {
-            Some(ref separator) => separator,
-            _ => "",
-        };
-
-        let group_separator = match self.separator {
-            Some(ref separator) => separator,
-            _ => "_",
-        };
+        let separator = self.separator.as_deref().unwrap_or("");
+        let group_separator = self.separator.as_deref().unwrap_or("_");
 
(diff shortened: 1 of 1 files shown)


src/env.rs
```diff
@@ -83,35 +83,26 @@ impl Source for Environment {
...
-            Some(ref separator) => separator,
-            _ => "",
-        };
-
-        let group_separator = match self.separator {
-            Some(ref separator) => separator,
-            _ => "_",
-        };
+        let separator = self.separator.as_deref().unwrap_or("");
+        let group_separator = self.separator.as_deref().unwrap_or("_");
```
matthiasbeyer, line 87: This is nice, but it would require that we (and our users) would use rust 1.40.0 or newer.
So I would like to have a CI job for that version before I merge this piece.

I will see if I can get to it myself in the next days, just leaving this as a comment and self-reminder!

matthiasbeyer (approved):

matthiasbeyer:
Well, CI says no.

For the clippy lints, simply rebase to master please. The other things (the rustfmt failure) have to be fixed, of course!

eduardocanellas:
Rebased and formatted :+1:

matthiasbeyer (approved):

PR #217, opened 2021-07-29
Preserve map ordering

### LINKED ISSUE #216: Keep ordering of deserialized maps
Serde's yaml deserialization supports maintaining the order of map types (at least in yaml/json, haven't looked at other formats) by using `LinkedHashMap` instead of `HashMap`. Quickly perusing through `config-rs`, it looks like the `table` value uses `HashMap` and so will always lose the ordering of a map.

AFAICT, a fix to this would be as simple as changing `HashMap` to `LinkedHashMap` in the appropriate places, but since that type is exposed to the API, it could be considered a breaking change. Would the project be receptive to such a change?

dlo9:
Fixes #216 by using [`LinkedHashMap`](https://crates.io/crates/linked-hash-map) instead of `HashMap` to preserve the order of the map types in configs. A few things to note:

- Ordering is not preserved when using `hjson` because the current dependency breaks when using a `LinkedHashMap`. Once that crate [adds support for Serde 1.0](https://github.com/hjson/hjson-rust/pull/24), the feature can be re-enabled and everything should work. There is [an alternative](https://crates.io/crates/deser-hjson) `hjson` serde implementation, but it makes a [design descision](https://github.com/Canop/deser-hjson/tree/1fb13c996764a91334ccf8c4f7ff834bf89be616#why-only-a-derive-based-deserializer) incompatible with use in this project
- Merged configs will be order-preserving within each config -- that is, given the following configs in order:
  ```yaml
  # Config 1
  creator:
    name: John Smith
    username: jsmith
  ```

  ```yaml
  # Config 2
  creator:
    email: jsmith@localhost
    name: John Smith Jr.
  ```

  The resulting config will contain settings in the order found, preserving the order within each individual config only for new settings. Updates to settings is a lower config will have no effect on ordering:
  ```yaml
  # Merged config
  creator:
    name: John Smith Jr.
    username: jsmith
    email: jsmith@localhost
  ```

TODOs:
- [x] Overall, the community seems to be converging on use of [`indexmap`](https://crates.io/crates/indexmap) instead of `LinkedHashMap` for this use case. This is due to some improved performance, less unsafe code, and more active development (`LinkedHashMap` is in maintenance mode). I initially implemented this using `LinkedHashMap` because I wasn't aware of the other crate, but would like to use `index-map` before marking the PR ready
- [x] This current implementation would be a breaking change, but I believe can be made a non-breaking feature by [type aliasing](https://github.com/ron-rs/ron/blob/3984332e191b9eb1f12b2a11fc7dd92cf56387e9/src/value.rs#L129-L132). I would like to also make this change pre-merge
- [x] Check for over-replacement of `HashMap` -> `IndexMap`. I expect there are uses that don't need to be replaced because they don't impact the iteration order of the resulting config, but I haven't yet verified


matthiasbeyer:
So far this looks good to me. Take your time, of course! :+1: No need to rush things here :smile:

Cargo.toml
```diff
@@ -44,3 +46,6 @@ tokio = { version = "1", features = ["rt-multi-thread", "macros", "fs", "io-util
 warp = "0.3.1"
 futures = "0.3.15"
 reqwest = "0.11.3"
+# Workaround to activate non-default features for tests:
+# https://github.com/rust-lang/cargo/issues/2911
+config = { path = ".", features = ["preserve_order"] }
```
dlo9, line 51: This may not want to be kept -- it's not a default feature to keep this an unbreaking change, but I wanted the tests to run with it enabled to prove it works

matthiasbeyer: I think it would be best to alter the actions step to simply use `test --all-features` and be done with it?

dlo9: Done :+1:

dlo9:
@matthiasbeyer any idea when you'll have a chance to look at this?

matthiasbeyer (approved):

matthiasbeyer:
Sorry for the delay :neutral_face:

dlo9:
@matthiasbeyer No worries --  you've been very responsive, I just wanted to make sure it wasn't forgotten :smile:

matthiasbeyer:
So, running rustfmt and adding a seperate commit is not nice history-wise.

But, there's help! Please try the following in your local repository on your PR branch

```bash
git rebase --strategy-option=theirs --exec "cargo fmt && git add --all && git commit --amend --no-edit" <this repository>/master
```

then, make sure everything is fine by comparing your local branch with your pushed branch, something like this:

```bash
git diff HEAD..<this remote>/master
```

(btw, why are you PRing from `master` :laughing: ?)

And if there's no diff, you can push the changes to this PR. Maybe the last commit ("Run rustfmt") is not necessary anymore... I tried this locally with you branch and the commit only contains one change, a removed import. Thus it might be good to just change the commit message :smile: 

Please try this! If you think you messed up I am here to help (and I have a copy of this PR locally, I can always simply roll back if you accidentially messed up :smile: )

dlo9:
I had assumed you would squash-merge this, since the commit history in this PR has some churn and isn't terribly useful IMO. As a general rule I don't rebase after opening a PR so that I don't make the review confusing (unless coordinated with reviewers of course).

Though if you'd like me to fix up the history I'm more than happy to, would you prefer that or squashing?

matthiasbeyer:
I never squash merge because that kills history. So yes, please, rebase this 😊

dlo9:
@matthiasbeyer friendly ping after the rebase -- I didn't remove any history except for the formatting commit

matthiasbeyer (approved):

matthiasbeyer:
Thanks for your contribution! I hope you come back for more! :heart: :laughing:

PR #218, opened 2021-07-30
Add Dhall support

sphw:
This PR adds Dhall support using [serde_dhall](https://github.com/Nadrieril/dhall-rust), a native Rust implementation of Dhall. Dhall is a functional configuration language designed to make composition of configuration files easier.

closes https://github.com/mehcode/config-rs/issues/123

### DIFF
--- a/src/file/format/dhall.rs
+++ b/src/file/format/dhall.rs
@@ -1,0 +1,55 @@
+use std::collections::HashMap;
+use std::error::Error;
+
+use crate::{
+    error::Unexpected,
+    value::{Value, ValueKind},
+    ConfigError,
+};
+
+pub fn parse(
+    uri: Option<&String>,
+    text: &str,
+) -> Result<HashMap<String, Value>, Box<dyn Error + Send + Sync>> {
+    let value = from_dhall_value(uri, serde_dhall::from_str(text).parse()?);
+    match value.kind {
+        ValueKind::Table(map) => Ok(map),
+        ValueKind::Nil => Err(Unexpected::Unit),
+        ValueKind::Boolean(value) => Err(Unexpected::Bool(value)),
+        ValueKind::Integer(value) => Err(Unexpected::Integer(value)),
+        ValueKind::Float(value) => Err(Unexpected::Float(value)),
+        ValueKind::String(value) => Err(Unexpected::Str(value)),
+        ValueKind::Array(value) => Err(Unexpected::Seq),
+    }
+    .map_err(|err| ConfigError::invalid_root(uri, err))
+    .map_err(|err| Box::new(err) as Box<dyn Error + Send + Sync>)
+}
+
+fn from_dhall_value(uri: Option<&String>, value: serde_dhall::SimpleValue) -> Value {
+    match value {
+        serde_dhall::SimpleValue::Num(num) => match num {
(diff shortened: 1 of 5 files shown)


src/file/format/dhall.rs
```diff
@@ -0,0 +1,49 @@
...
+use std::error::Error;
+
+use crate::value::{Value, ValueKind};
+
+pub fn parse(
+    uri: Option<&String>,
+    text: &str,
+) -> Result<HashMap<String, Value>, Box<dyn Error + Send + Sync>> {
+    // Parse a Dhall object value from the text
+    // TODO: Have a proper error fire if the root of a file is ever not a Table
```
matthiasbeyer, line 11: I guess we can resolve this right away, can't we?

matthiasbeyer (changes requested):
Hi,

Awesome, thanks for your contribution!

In general this looks good. The TODO item could be fixed right away, right?

Also (I tested this because I was not 100% sure whether it would work), the `from_dhall_value()` function does not need to clone.

Here is a patch for that: https://github.com/matthiasbeyer/config-rs/commit/7a163a74ae54f524e486ee1a8c173df800a14709
And here's another patch you could apply: https://github.com/matthiasbeyer/config-rs/commit/91dafd157021bc280750d4d5f2c20a8427687f2b

I guess you can just `git pull --ff-only https://github.com/matthiasbeyer/config-rs/ pr-218` if you want to apply these :smile:

matthiasbeyer:
Plus, I messed up the formatting (thank god we have CI for that), here's the patch that fixes it: https://github.com/matthiasbeyer/config-rs/commit/00ed90b908b3fca4cce7c4c2385b5a51df32bef9

You can of course still use the `git pull --ff-only` line from above, the patch was pushed to this branch and a `git rebase --autosquash -i master` will squash the fixup away :smile: 

Sorry for messing this up...

src/file/format/dhall.rs
sphw: I shamelessly stole this from the JSON impl, including the comment. I figured having consistent error handling was important.

What is the best type of error to throw in this case?

sphw: Actually, I noticed that the error is properly handled in the JSON5 implementation. I'll use that as a model

sphw:
@matthiasbeyer Thanks for the timely and thorough review

I integrated all your requested changes and comments, and then squashed them into a single commit. I also fixed the error handling case you mentioned above. 

Let me know if you have any other comments

matthiasbeyer (approved):
Looks good.

No need to squash things into one commit, though. That only makes reviewing harder. For the next time, then... :laughing:

MGlolenstine:
I have published a [PR](https://github.com/sphw/config-rs/pull/1) that should fix the problems during the build.
All tests do succeed now.

polarathene:
@matthiasbeyer it looks like this PR is going stale too? The build PR referenced last in Dec 2021 was merged, and no remaining concerns with this PR for being merged?

matthiasbeyer:
Yep so there's conflicts and this PR needs rebasing.

polarathene:
> Yep so there's conflicts and this PR needs rebasing.

Done see #466

sphw:
@polarathene I'm closing in favor of your PR, since this has been sitting for almost 2 years now lol. Thanks so much for helping to push it over the finish line.

PR #219, opened 2021-08-01
Introduce `Format` trait

szarykott:
This pull request introduces a `Format` trait (with its companion `FileExtensions` to satisfy legacy behavior).

Having such a trait allows users of the library to use custom/proprietary/less known formats with config-rs without the need to alter library code, which is not only a lengthy process, but might be impossible for some formats. What is more, having it could facilitate creating sources other than `File` that operate on the notion of Format, but are independent of a file concept (network sources).

This pull request does not alter anything functionally speaking, but adds more genericness to already existing infrastructure. It makes the library more extensible.

I hope you like the idea!


szarykott:
It seems I will have to check Clippy on nightly.

examples/custom_format/main.rs
```diff
@@ -0,0 +1,55 @@
...
+
+        Ok(result)
+    }
+}
+
+// As crazy as it seems for config sourced from a string, legacy demands its sacrifice
+// It is only required for File source, custom sources can use Format without caring for extensions
+static MY_FORMAT_EXT: Vec<&'static str> = vec![];
+impl FileExtensions for MyFormat {
+    fn extensions(&self) -> &Vec<&'static str> {
```
matthiasbeyer, line 52: So if we have a `&'static str` here, wouldn't it be possible to remove the (heap allocated, if I'm not mistaken) `Vec` here as well?

Maybe with something like a `fn extensions(&self) -> impl Iterator<Item = &'static str>` kind of signature?

src/file/extension.rs
```diff
@@ -0,0 +1,14 @@
+/// Describes an entity that can return file extensions.
+///
+/// Meant primarily for [`Format`](crate::Format) trait.
+/// Since [`Format`](crate::Format) only describes certain internal encoding, for instance JSON or Yaml
+/// it is not necessarily bound to file extension name.
+///
+///In networking context JSONs are used without file extensions.
```
matthiasbeyer, line 7: nit: missing leading space

src/file/format/mod.rs
```diff
@@ -93,18 +96,18 @@ lazy_static! {
 impl FileFormat {
     // TODO: pub(crate)
     #[doc(hidden)]
-    pub fn extensions(self) -> &'static Vec<&'static str> {
+    pub fn extensions(&self) -> &'static Vec<&'static str> {
```
matthiasbeyer, line 99: There's a `TODO: pub(crate)` here, that could be resolved right away, right? :smile:

src/file/mod.rs
```diff
@@ -1,35 +1,46 @@
...
 
     /// A required File will error if it cannot be found
     required: bool,
 }
 
-impl File<source::string::FileSourceString> {
-    pub fn from_str(s: &str, format: FileFormat) -> Self {
+impl<F> File<source::string::FileSourceString, F>
+where
+    F: Format + FileExtensions + 'static,
```
matthiasbeyer, line 41: I wonder how the `'static` bound behaves in production code then... :thinking:

matthiasbeyer (changes requested):
First of all: Sorry for not coming back to this, I had a lot on my mind lately (and still have, so another roundtrip might take some time again, too... sorry about that)!

So for the PR: Well, this is quite the change, isn't it? I am not yet sure how this all will play out in production code later... Maybe I can find some time to actually try to move my production codebase to use this changeset to see how it works out - but that might take a fair bit of time.

I hope you don't mind this living for a bit longer... :pensive:

examples/custom_format/main.rs
szarykott: Good point with the `impl`, more flexible! Static lifetime should be enough, file extensions are quite ... static.

src/file/format/mod.rs
szarykott: Sure, I'll see what I can do

szarykott:
@matthiasbeyer I'll try to answer your concerns about production here.

Previously `File` source accepted only well-defined structs (so by definition none could use anything else). Those structs (one struct to be precise) all implement introduced traits, hence code should be backwards compatible - just see that I need not modify any test to make it compile and work. 

As I see it, it only allows custom formats to be defined and integrated into library by users without breaking backwards compatibility.

src/file/mod.rs
szarykott: It only disallows `F` implementations to have borrowed fileds, it can be easily worked around with a simple `clone` or `Rc`. Unless somebody adds sources to config in a hot loop in a critical path, this should be fine.

examples/custom_format/main.rs
szarykott: To avoid heap allocation, I need to go with `&'static [&'static str]`.

Why?
* `impl Iterator<Item = &'static str>` is forbidden in traits 
* to define associated type Iterator<Item = &'static str> I'd need to return a reference to a slice and I think that to pass lifetime to it I'd need GAT.

src/file/format/mod.rs
szarykott: Required some changes in the example, but it is all pub (crate) now.

szarykott:
Hey, @matthiasbeyer I pushed a lot of changes to make things green.

What do you think about this PR now?

szarykott:
@matthiasbeyer I'd like to give it one more look myself, but a review from somebody else would not be a bad idea. Are there people willing to do it?

szarykott:
@matthiasbeyer what do you think?

matthiasbeyer (approved):

matthiasbeyer:
:partying_face:

PR #223, opened 2021-08-21
Make examples recognized by cargo

SkiFire13:
#222 fixed the imports in the examples, however it didn't actually fix CI. The reason is that cargo doesn't recognize crate examples, only single files or a `main.rs` inside a single folder (not nested like in a crate). This PR fixes this by changing the file structure in such a way that cargo recognized them. Note that:

- I `cargo fmt`s the examples because now that cargo can see them it can also format them. This wasn't the case before.
- I removed the `cargo check --examples` from CI because it is redundant: `cargo test` already checks them, but both commands require the file structure to be correct.
- I also updated the examples to use `ConfigBuilder` in order to fix the warnings. This might need some feedback here since I'm not 100% this is right.
- The config position relative to the `main.rs` is not accurate anymore unfortunately.


matthiasbeyer:
Grml... I should've checked this more carefully! Thanks a lot for fixing this up!

examples/global/main.rs
```diff
@@ -0,0 +1,27 @@
...
+use std::sync::RwLock;
+
+lazy_static! {
+    static ref SETTINGS: RwLock<Config> = RwLock::new(Config::default());
+}
+
+fn try_main() -> Result<(), Box<dyn Error>> {
+    // Set property
+    #[allow(deprecated)]
+    SETTINGS.write()?.set("property", 42)?;
```
SkiFire13, line 17: Ah, I forgot to add that I wasn't really sure what should have gone where, so I just suppressed the warning. This should definitely be replaced with a better solution. It would be a bit weird if examples used deprecated stuff.

examples/watch/main.rs
```diff
@@ -44,6 +47,7 @@ fn watch() {
         match rx.recv() {
             Ok(DebouncedEvent::Write(_)) => {
                 println!(" * Settings.toml written; refreshing configuration ...");
+                #[allow(deprecated)]
                 SETTINGS.write().unwrap().refresh().unwrap();
```
SkiFire13, line 51: Same here

examples/global/main.rs
matthiasbeyer: Yes, it definitively is.

SkiFire13:
I've removed those `#[allow(deprecated)]`, and although I don't really like the end result I don't think there are better ways with the current state of the crate.

matthiasbeyer:
Superceeded by #253

Issue #224, opened 2021-08-23
Unchecked integer range


BratSinot:
Greetings!

`config` lib is casting all integers like explicit type conversion through `as`. Such behaviour by design, or this is an error?

```rust
use config::Config;

fn main() {
    let mut conf = Config::new();

    conf.merge(config::File::from_str(
        r#"
        [server]
        port = 66000
        "#,
        config::FileFormat::Toml,
    ))
    .unwrap();

    let port: u16 = conf.get("server.port").unwrap();

    println!("{}", port); // output is 464
}
```

BratSinot:
Maybe duplicate of this #93.

matthiasbeyer:
Yes, this is actually a limitation.

The problem is here that the format (toml in this case, but others apply here as well) does not define the datatype (besides "Number"). By specifying `let port: u16` in your code, the value is inferred to be `u16` and the library just assumes that the number is actually a 16 bit unsigned integer.

We cannot do much here, on the library side of things, actually (or I'm wrong and someone comes up with something clever...).

We need you to know about your data, so you have to make sure that you ask for a `u32` here and all will be fine I guess.

BratSinot:
> Yes, this is actually a limitation.
> 
> The problem is here that the format (toml in this case, but others apply here as well) does not define the datatype (besides "Number"). By specifying `let port: u16` in your code, the value is inferred to be `u16` and the library just assumes that the number is actually a 16 bit unsigned integer.
> 
> We cannot do much here, on the library side of things, actually (or I'm wrong and someone comes up with something clever...).
> 
> We need you to know about your data, so you have to make sure that you ask for a `u32` here and all will be fine I guess.

Thanks for answer!
Now I'm using this workaround:
```rust
    let port: u16 = conf
        .get_int("server.port")
        .unwrap()
        .try_into() // here
        .unwrap();
```
or this:
```rust
    let port = conf
        .get_string("server.port")
        .unwrap()
        .parse::<u16>()
        .unwrap();
```

matthiasbeyer:
Okay, I'm closing this therefore, feel free to ping me if you have more questions on the matter! :smile:

PR #225, opened 2021-08-23
Add tests for wrapping integers

matthiasbeyer:
Context: #224


matthiasbeyer:
Can we merge this? I think we can, but maybe someone else has some ideas here?

PR #230, opened 2021-09-26
Remove support for hjson

### LINKED ISSUE #198: Please consider switching upstream implementations to reduce repetitive dependencies
Recently I was checking the project structure and found that config repeatedly depends on multiple versions of serde:

image

So I found an issue (discuss), see: https://github.com/hjson/hjson-rust/issues/6

plz consider switching serder-hjson to deser_hjson, or switching serder-json to v0.8

matthiasbeyer:
The serde-hjson crate is not maintained anymore and this feature is
actually causing pain in packaging even.

Thus, remove it.

---

Closes #198


matthiasbeyer:
@dlo9 you wrote the [line that fails clippy](https://github.com/mehcode/config-rs/pull/230/checks?check_run_id=3713021178#step:5:115) here. Am I just stumbling on a bug because this field is not used anywhere?

dlo9:
@matthiasbeyer My change on those lines was just changing `HashMap` to the new `Map` type alias, there was no substantial change to those fields. Nightly clippy fails with the same errors when run on the commit prior to my PR.

AFAICT these are new checks that nightly clippy added recently, and which are completely valid, but just haven't previously been run on this codebase. Most of these are because #207 added the `AsyncConfigBuilder` but doesn't have an `impl` for it, so its fields aren't used anywhere.

matthiasbeyer:
I guess I'll merge this then, the fixes for the clippy warnings should go into a seperate PR anyways.

PR #234, opened 2021-10-02
Test features individually

### LINKED ISSUE #233: unit tests doesn't gate features properly
When running the unit test suite per feature (as debians build system does) you get compile errors:

For example:

```
$ cargo test --no-default-features --features yaml
   Compiling config v0.11.0 (/tmp/config-rs)
error[E0599]: no variant or associated item named `Json` found for enum `FileFormat` in the current scope
  --> examples/async_source/main.rs:39:33
   |
39 |             format: FileFormat::Json,
   |                                 ^^^^ variant or associated item not found in `FileFormat`

For more information about this error, try `rustc --explain E0599`.
error: could not compile `config` due to previous error
warning: build failed, waiting for other jobs to finish...
error: build failed

```

matthiasbeyer:
Closes #233 

This PR aims to resolve #233 - the issue I have right now is that the doctests are written with the `json` feature, but they fail if I run CI with `cargo test --no-default-features --feature ini` for example.

If someone knows how to enable the doctests only when `json` is enabled, that'd be awesome! :laughing:


epage:
Appears to be done, see https://github.com/rust-cli/config-rs/blob/5f308736784460440f438b52296edc2b8e650102/.github/workflows/ci.yml#L53

PR #236, opened 2021-10-02
Fix panic: Unwrap()ing a f64::parse() should not happen

### LINKED ISSUE #231: panic when parsing yaml
Full error message:

```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: ParseFloatError { kind: Invalid }', /home/capitol/.cargo/registry/src/github.com-1ecc6299db9ec823/config-0.11.0/src/file/format/yaml.rs:37:67

```

Program to reproduce the error:

```
use config::{Config, File, FileFormat};

fn main() {
    let s = vec![0x2e, 0x6e, 0x61, 0x6e, 0x00, 0x08, 0x00, 0x00, 0x00, 0x30, 0x78, 0x22,
                 0x4c, 0x4c, 0x2e, 0x6e, 0x61, 0x6e, 0x32, 0x32, 0x32, 0x32, 0x32, 0x34,
                 0x5c, 0x66, 0x41, 0x38, 0x38, 0x32];

    let _ = Config::new().merge(File::from_str(std::str::from_utf8(&s).unwrap(), FileFormat::Yaml));
}

```

matthiasbeyer:
This fixes a panic where a f64 failed to be parsed, which caused
end-user software to crash.

This patch fixes the problem.

Closes #231

@alexanderkjall can you test this patch whether your issue is reproducible?
I'm not sure whether I can backport this to 0.11.x because the codebase changed a lot since, but I will give it a try if it fixes your issue.


alexanderkjall:
This seems to fix the problem :)

No need to backport it I think, better to spend time moving forward :)

PR #242, opened 2021-10-21
feat: allow providing the environment for testing

### LINKED ISSUE #241: Allow providing a custom map for Environment
Using `Environment` can sometimes be a bit tricky. Instead of writing unit tests, setting and restoring actual system env-vars, it would be great if an `Environment` instance could be created, passing it the source of the variables, which then could also be a user provided map.

ctron:
closes #241

### DIFF
--- a/src/env.rs
+++ b/src/env.rs
@@ -26,6 +26,42 @@ pub struct Environment {
 
     /// Parses booleans, integers and floats if they're detected (can be safely parsed).
     try_parsing: bool,
+
+    /// Alternate source for the environment. This can be used when you want to test your own code
+    /// using this source, without the need to change the actual system environment variables.
+    ///
+    /// ## Example
+    ///
+    /// ```rust
+    /// # use config::{Environment, Config};
+    /// # use serde::Deserialize;
+    /// # use std::collections::HashMap;
+    /// # use std::convert::TryInto;
+    /// #
+    /// #[test]
+    /// fn test_config() -> Result<(), config::ConfigError> {
+    ///   #[derive(Clone, Debug, Deserialize)]
+    ///   struct MyConfig {
+    ///     pub my_string: String,
+    ///   }
+    ///
+    ///   let source = Environment::default()
+    ///     .source(Some({
+    ///       let mut env = HashMap::new();
+    ///       env.insert("MY_STRING".into(), "my-value".into());
+    ///       env
+    ///   }));
+    ///
+    ///   let config: MyConfig = Config::builder()
+    ///     .add_source(source)
+    ///     .build()?
+    ///     .try_into()?;
+    ///   assert_eq!(config.my_string, "my-value");
+    ///
+    ///   Ok(())
+    /// }
+    /// ```
+    source: Option<Map<String, String>>,
 }
 
 impl Environment {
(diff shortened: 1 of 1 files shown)


conradludgate:
This intrigued me so I made my own alternative implementation. It's a bit ugly due to requiring a few dyn objects, but it's a generic version of this feature.

https://github.com/conradludgate/config-rs/commit/2ff2183ac1b8798b18417d814a89fa4510e6a062

src/env.rs
```diff
@@ -26,6 +26,9 @@ pub struct Environment {
 
     /// Parses booleans, integers and floats if they're detected (can be safely parsed).
     try_parsing: bool,
+
+    /// Alternate source for environment. This can be used for testing.
+    source: Option<Map<String, String>>,
```
matthiasbeyer, line 31: I'm not sure whether you mean testing of this crate or testing when using this crate.

If the former, this should be `#[cfg(test)]` IMO! Otherwise it would be nice to have an example here,... or some words how it can be used!

matthiasbeyer (approved):
The general diff looks good to me, but I'd like to clearify the mentioned point before merging! :+1:

src/env.rs
ctron: The latter. I added a few more words and an example. I hope this helps.

src/env.rs
```diff
@@ -82,7 +123,12 @@ impl Source for Environment {
             .as_ref()
             .map(|prefix| format!("{}{}", prefix, group_separator).to_lowercase());
 
-        for (key, value) in env::vars() {
+        let source: Box<dyn Iterator<Item = (String, String)>> = match &self.source {
+            Some(source) => Box::new(source.clone().into_iter()),
+            None => Box::new(env::vars()),
+        };
+
+        for (key, value) in source {
```
conradludgate, line 131: I don't like this. I think it would be better to move the body of the for loop into a closure, then in the match you'd do something like this instead:

```rust
match &self.source {
    Some(source) => source.cloned().into_iter().for_each(body),
    None => env::vars().for_each(body),
}
```

ctron: You are right, that looks way cleaner. Thanks for sharing the improvement!

src/env.rs
```diff
@@ -122,6 +163,11 @@ impl Source for Environment {
             };
 
             m.insert(key, Value::new(Some(&uri), value));
+        };
+
+        match &self.source {
+            Some(source) => source.clone().into_iter().for_each(collector),
```
matthiasbeyer, line 169: do we actually need to `clone()` here? AFAICS, the collector could use references for the key-value-pair, couldn't it? Or would that conflict with the `env::vars()` iterator, because that one returns `(String, String)`?

ctron: So `env::vars()` provides `(String,String)`, and some paths in the (now) closure consume the string value.

The `self.source` does not get moved, as it is part of `self`. I tried with a closure of `(AsRef<str>, Into<String>)`, however, this doesn't work because both are not `Sized`. That would require introducing a new function with type arguments. As this is code intended for testing, I think having the `clone` is ok. Besides, the `env::vars()` also clones, but hides this.

ctron:
I this one ok to be merged? Currently this is an issue for us, as we need to set actual env-vars in tests, and that causes flaky tests.

We could synchronize them, run tests sequentially, or use this feature instance. I would hope for the latter.

matthiasbeyer (approved):

matthiasbeyer:
:partying_face:

PR #247, opened 2021-11-13
feat: allow for current_dir failure

conradludgate:
We have a potential use-case where a config is loaded while in a directory that doesn't exist (see https://github.com/ellie/atuin/issues/195). This PR makes the code a bit more accepting of whether the `env::current_dir` command fails, falling back on the filename as the `uri`.

### DIFF
--- a/src/file/source/file.rs
+++ b/src/file/source/file.rs
@@ -103,11 +103,10 @@ where
         let (filename, format) = self.find_file(format_hint)?;
 
         // Attempt to use a relative path for the URI
-        let base = env::current_dir()?;
-        let uri = match path_relative_from(&filename, &base) {
-            Some(value) => value,
-            None => filename.clone(),
-        };
+        let uri = env::current_dir()
+            .ok()
+            .and_then(|base| path_relative_from(&filename, &base))
+            .unwrap_or_else(|| filename.clone());
 
         // Read contents from file
         let mut file = fs::File::open(filename)?;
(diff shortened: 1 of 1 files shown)


src/file/source/file.rs
```diff
@@ -103,16 +103,13 @@ where
...
-            Some(value) => value,
-            None => filename.clone(),
-        };
+        let uri = env::current_dir()
+            .ok()
+            .and_then(|base| path_relative_from(&filename, &base))
+            .unwrap_or_else(|| filename.clone());
 
         // Read contents from file
-        let mut file = fs::File::open(filename)?;
```
matthiasbeyer, line 113: This change does not belong to the other change you provided - please split it into a seperate commit!

matthiasbeyer (changes requested):
Overall the diff looks good, except for the one thing I annotated.

CI fails because of an example, which should be fixed as well :smile:

conradludgate:
The CI failure looks like a network error, not a result of these changes

conradludgate:
The newest version of warp (`0.3.2`) included a new version of `tungsenite` which makes use of const generics. Pinning it to `0.3.1` fixes that

matthiasbeyer:
Sorry for taking so long, this week was really busy for me.

Did you test this change with your codebase? Because right above your change, `FileSourceFile::find_file()` is called, which calls `env::current_dir()?` right away, which would result in the same thing you're trying to fix here :laughing:

conradludgate:
Good point. I'll test it and see if it's all working. I'll then rebase after

matthiasbeyer (approved):

Issue #162, opened 2020-12-30
Value::try_into makes use of TryInto difficult


djmitche:
I'd like to write some conversions for my own structs from instances of Value, and `std::convert::TryFrom` / `TryInto` are good choices for this.  However, the existence of a `try_into` method on the struct, unrelated to these traits, shadows the trait method.

Could the `try_into` method be renamed into `try_deserialize` or something of the sort?

matthiasbeyer:
IMO, the `try_from` function should be reimplemented using the `TryFrom` trait anyways.

I think this is a good idea!

matthiasbeyer:
Fixed in [config-rs-maint](https://git.sr.ht/~matthiasbeyer/config-rs-maint/) as of [master def3702fac13605af303b8319af15943cc7ff8e5](https://git.sr.ht/~matthiasbeyer/config-rs-maint/commit/def3702fac13605af303b8319af15943cc7ff8e5).

conradludgate:
Sadly I don't think it's possible to implement it using TryFrom for all T because it conflicts.

I'm not _quite_ sure what with, but I think it's because a type could implement `From<Value>` and `Deserialize` individually downstream and that would have conflicting implementations

conradludgate:
Should we just go with `try_deserialize()` for now?

djmitche:
that sounds good to me :)

I'm confused, though, because it looks like @matthiasbeyer already did this?

conradludgate:
> I'm confused, though, because it looks like @matthiasbeyer already did this?

Looking through their old maint fork, they only had this
https://git.sr.ht/~matthiasbeyer/config-rs-maint/tree/master/item/src/value.rs#L154. No TryFrom that I can find

matthiasbeyer:
Yeah I even tried to implement the `TryInto` trait, but it was not possible when I tried... But having a function named `try_into()` is counter-intuitive, because it suggestes that the trait is there... which it is not. So I guess `try_deserialize()` it is, right?

conradludgate:
I would say that's a consensus reached

danieleades:
looks like this is already merged. can this be closed?

matthiasbeyer:
Yes! Thanks for pinging me!

PR #251, opened 2021-11-22
rename try_into to try_deserialize to avoid confusion

conradludgate:
For #162

### DIFF
--- a/examples/glob/src/main.rs
+++ b/examples/glob/src/main.rs
@@ -16,7 +16,7 @@ fn main() {
 
     // Print out our settings (as a HashMap)
     println!("\n{:?} \n\n-----------",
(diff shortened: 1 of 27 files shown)


matthiasbeyer (changes requested):
Diff looks good!
Please squash the commits into one and use an appropriate commit message!

matthiasbeyer:
Yeah, please remove the `feat:` prefix from the commit message, then I'll merge!

conradludgate:
Bah, I'm a sucker for conventional commits. I'll change it up though

matthiasbeyer (approved):
I don't think conventional commits provide any value,... we can discuss this if you want, but here's not the place! :smile: :heart:

PR #252, opened 2021-11-23
update tests not use deprecated methods

conradludgate:
I was tired of the lint warnings in my editor :sweat_smile:. These tests don't seem to be for that specific legacy functionality so these tests shouldn't be considered regressions


matthiasbeyer (approved):
Yippie! I wanted to do this myself for a long time but had no time!

Thanks so much for this contribution!

Issue #176, opened 2021-03-16
Storing config back into file


smallB007:
Hi, 
I was able to successfully use your crate in order to read config from a file.
Unfortunately, I was not able to write changed config back to the same file. Could somebody provide me with a hint how to go about it?
Thank you

matthiasbeyer:
I guess this is not possible natively with this library.

What you could do, though, is deserialize your configuration into a normal rust type and then serialize that back to a toml/json/ini/whatever and (over)write the file.

smallB007:
Hey, thanks for answer, but... How such an essential feature is not there is simply beyond me...
I mean, so an application can read the configuration, change it and not being able to save it? Seems like awful overlook to me, tbh...

matthiasbeyer:
You're more than welcome to contribute such a feature!

Though, I guess, the feature is simply not possible because the format-implementing crates (serde_json, toml, yaml) do not support serializing to the format while keeping the order and comments, e.g. think toml:

```toml
# this is a value
value = 5

# this is something else
foo = "helloworld"
```
Which would be, when serialized back from rust (values not changed):

```toml
foo = "helloworld"
value = 5
```

So both comments and order would be lost. I'd bet that's true for the other backends (formats), too, at least for some.

If you find that my statement is false, I'd love to get a PR for this crate :smile:

smallB007:
Hey, thanks for quick reply.
I actually think that loading, changing and saving config is possible. Literally thousands (if not millions) of applications are doing it.
Can't promise PR for that, but who knows, if I find time I'll certainly have a go at it.
Best regards

matthiasbeyer:
> Literally thousands (if not millions) of applications are doing it.

Sure! I just am curious about whether this is possible with the libraries we use in this crate!

> Can't promise PR for that, but who knows, if I find time I'll certainly have a go at it.

Sure! I'll keep this issue open for others to step up or just to see that there's a discussion about that topic.

conradludgate:
So the comment/order preserving features [_do_ exist for toml](https://docs.rs/toml_edit/0.9.1/toml_edit/index.html), but it's rather specific.

There's two ideas I think are clashing here.

1. Some applications might maintain their own config through a GUI, in which case the comments/order doesn't matter to the user.
2. Then there's applications like servers or CLI applications where the config is only modified by the user.

I think the goal of this lib is to mostly support the latter. If the config is maintained by the application, I think having your own JSON serialize/deserialize steps is much simpler

[...]

PR #254, opened 2021-11-29
Add note about writing back config

matthiasbeyer:
After the discussion in #176 this seems like a good idea.


smallB007:
> After the discussion in #176 this seems like a good idea.

You are welcome...

PR #255, opened 2021-12-03
feat: env contains list of strings

### LINKED ISSUE #117: environment variables list?
is it possible to parse an env var into something like `Vec<String>`? I get this error for any formatting I try:

```
thread 'main' panicked at 'somethings wrong with the settings: invalid type: string "1,2,3", expected a sequence', src/libcore/result.rs:999:5
```

the error to me suggest that it is possible, but I could just be reading things wrong.

simon-an:
Signed-off-by: simon-an <26556185+simon-an@users.noreply.github.com>

fixes #117

### DIFF
--- a/src/env.rs
+++ b/src/env.rs
@@ -5,6 +5,7 @@ use crate::map::Map;
 use crate::source::Source;
 use crate::value::{Value, ValueKind};
 
+#[must_use]
 #[derive(Clone, Debug, Default)]
 pub struct Environment {
     /// Optional prefix that will limit access to the environment to only keys that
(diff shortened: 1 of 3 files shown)


matthiasbeyer (changes requested):
Codewise this looks good, but please fix the commit message and use a real signoff line! :smiley:

simon-an:
ups. :) Done.

simon-an:
I noticed, that with my first implementation one have to decide what to use. Strings or lists of strings. So i added an option which allows adding the keys of environment variables to be whitelisted for parsing.

simon-an:
Any chance we get this merged?
If you want to have a more sophisticated approach (instead of adding keys the parser should use), tell me in what direction to go and i will implement it.

matthiasbeyer (approved):

simon-an:
@matthiasbeyer Hi, can we merge this now? checks are green and branch is up to date.

matthiasbeyer:
Yeah I think so. Please rebase the branch so the merge commits vanish! :+1:

src/env.rs
```diff
@@ -80,39 +87,55 @@ impl Environment {
...
 
     /// Note: enabling `try_parsing` can reduce performance it will try and parse
     /// each environment variable 3 times (bool, i64, f64)
-    #[must_use]
     pub fn try_parsing(mut self, try_parsing: bool) -> Self {
         self.try_parsing = try_parsing;
         self
     }
 
-    #[must_use]
```
matthiasbeyer, line 115: Why did you remove these annotations?

matthiasbeyer (changes requested):

src/env.rs
simon-an: I have put it on the struct instead of the methods.  So they don't need to be on every method.
https://doc.rust-lang.org/reference/attributes/diagnostics.html#the-must_use-attribute

matthiasbeyer (approved):

PR #257, opened 2021-12-26
remove explicit 'extern crate' calls

danieleades:


matthiasbeyer (changes requested):
Can you please split the commit into several so that we have one commit per type of change?

matthiasbeyer (approved):

matthiasbeyer (approved):

PR #258, opened 2021-12-26
add dependabot config

danieleades:


matthiasbeyer:
Do we need any special config for this repository for this change to have any effect?

danieleades:
> Do we need any special config for this repository for this change to have any effect?

negative, it should Just Work.

PR #260, opened 2021-12-26
small refactors of 'msrv' workflow

danieleades:


matthiasbeyer (changes requested):
Can you please split the commit into several so that we have one commit per type of change?

danieleades:
> Can you please split the commit into several so that we have one commit per type of change?

I feel like the changes in this PR are fairly atomic already. How do you see this being broken down?

matthiasbeyer:
One change for the `minimal` key and one for the other change,... If they are too tightly related in your opinion, at least a commit message that explains what this change actually does would be really nice.

Either way, thanks for your contribution! :heart:

danieleades:
> One change for the `minimal` key and one for the other change,... If they are too tightly related in your opinion, at least a commit message that explains what this change actually does would be really nice.
> 
> Either way, thanks for your contribution! heart

done. let me know if you're happy with those

matthiasbeyer (approved):

PR #262, opened 2021-12-26
ensure clippy runs on all targets

danieleades:
note that there are clippy false positives due to https://github.com/rust-lang/rust-clippy/issues/8140

### DIFF
--- a/.github/workflows/msrv.yml
+++ b/.github/workflows/msrv.yml
@@ -136,7 +136,7 @@ jobs:
         uses: actions-rs/cargo@v1
         with:
           command: clippy
-          args: -- -D warnings
+          args: --all-targets --all-features -- -D warnings
 
   check-examples:
     name: Check examples
--- a/examples/glob/main.rs
+++ b/examples/glob/main.rs
@@ -7,14 +7,12 @@ fn main() {
     // Option 1
     // --------
     // Gather all conf files from conf/ manually
-    let mut settings = Config::default();
-    settings
+    let settings = Config::builder()
         // File::with_name(..) is shorthand for File::from(Path::new(..))
-        .merge(File::with_name("examples/glob/conf/00-default.toml"))
-        .unwrap()
-        .merge(File::from(Path::new("examples/glob/conf/05-some.yml")))
-        .unwrap()
-        .merge(File::from(Path::new("examples/glob/conf/99-extra.json")))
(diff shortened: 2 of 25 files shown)


matthiasbeyer (changes requested):
Can you please split the commits so that we have only one commit per type of change?

matthiasbeyer (approved):

danieleades:
note that there is some remaining use of deprecated code in the examples. It's not clear to me how these use cases should be supported using the builder API. I've *allowed* these for now, but they should be addressed in a future PR

matthiasbeyer:
You're adding `#[must_use]` f7899fd in and removing it again in 48f49c6 - was this by accident? :laughing:

danieleades:
> You're adding `#[must_use]` [f7899fd](https://github.com/mehcode/config-rs/commit/f7899fd2dbefb311357af9f00af17f4b47fcece4) in and removing it again in [48f49c6](https://github.com/mehcode/config-rs/commit/48f49c684a266b8ab1a2353de74bdcf6c9d1ccd0) - was this by accident? laughing

suspect that snuck in during a rebase. I'll take a look

matthiasbeyer:
For the "global" example you can cherry-pick https://github.com/matthiasbeyer/config-rs/commit/1eff7b3a222aef41bff692377fc85fb1b00d760d to this PR if you want (or wait until the [actions](https://github.com/matthiasbeyer/config-rs/actions/runs/1633881795) succeed) :laughing:

danieleades:
> For the "global" example you can cherry-pick [matthiasbeyer@1eff7b3](https://github.com/matthiasbeyer/config-rs/commit/1eff7b3a222aef41bff692377fc85fb1b00d760d) to this PR if you want (or wait until the [actions](https://github.com/matthiasbeyer/config-rs/actions/runs/1633881795) succeed) laughing

this change doesn't look equivalent to the original to me. If i'm reading it right, the original example is *mutable* global config, while the updated version is *immutable* global config.

in any case, I think I'd rather leave it out of this PR. Don't want to over-complicate the review!

matthiasbeyer:
Okay!

matthiasbeyer:
Do you consider this review/merge ready? :smile:  I think I lost track a bit!

danieleades:
> Do you consider this review/merge ready? smile I think I lost track a bit!

yes, i'm not planning to do any more on this PR

matthiasbeyer (approved):

PR #263, opened 2021-12-26
bump deps

danieleades:


danieleades:
this is probably better handled by separate PRs per dependency (see #258)

it's the dev dependencies forcing the MSRV bump, so probably this isn't hugely worthwhile

danieleades:
closing in favour of dependabot PRs

PR #268, opened 2021-12-28
Add #[must_use] annotations

matthiasbeyer:
Clippy nightly fails the checks because it wants us to have a
`#[must_use]` annotation on functions that return `Self`.

So we add these annotations with this patch.

---

@danieleades I'm not sure whether you included this in your patchsets somewhere... :laughing:

### DIFF
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -181,6 +181,7 @@ impl ConfigBuilder<DefaultState> {
     /// Registers new [`Source`] in this builder.
     ///
     /// Calling this method does not invoke any I/O. [`Source`] is only saved in internal register for later use.
+    #[must_use]
     pub fn add_source<T>(mut self, source: T) -> Self
     where
         T: Source + Send + Sync + 'static,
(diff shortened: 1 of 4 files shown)


danieleades:
they're included in #262

note that i've made the types themselves `must_use` rather than the methods since this is appropriate for 'builder' structs. There is currently a false positive in clippy where the `must_use` suggestion fails to take into account that the type is `must_use`. see https://github.com/rust-lang/rust-clippy/issues/8140

matthiasbeyer:
@danieleades sorry if this is already included in your patchsets somewhere... I'll merge this because it makes CI happy right now, not because I don't like your work, okay? :laughing: :heart:

matthiasbeyer:
Ah, sorry, didn't see your comment before posting mine! Feel free to suggest any changes you see fit, of course!

danieleades:
> Ah, sorry, didn't see your comment before posting mine! Feel free to suggest any changes you see fit, of course!

how about i open a PR to move the `must_use` attributes to the type, rather than the methods, and then leave it as a draft until clippy is fixed upstream and the jobs pass?

matthiasbeyer:
Good idea!

PR #274, opened 2021-12-29
move 'must_use' attribute to struct for 'builder' types

danieleades:
note that this fails currently due to an upstream false positive in clippy.


matthiasbeyer:
Care to rebase this to latest master?

matthiasbeyer (approved):

PR #276, opened 2021-12-29
Refactor/tidy imports

danieleades:


examples/hierarchical-env/settings.rs
```diff
@@ -1,4 +1,5 @@
 use config::{Config, ConfigError, Environment, File};
+use serde_derive::Deserialize;
```
matthiasbeyer, line 2: Why is this import necessary if it wasn't before?

matthiasbeyer: Never mind... because you removed the `#[macro_use]` import in the `main.rs` file!

matthiasbeyer (approved):

PR #283, opened 2022-01-13
Add cron job

matthiasbeyer:
This patch adds a cron job to run cargo-check and cargo-clippy every
week, so that we are notified if the builds start failing because of
updated rustc/clippy versions if we did not have any commits in that
time.

---

Maybe someone wants to comment, so I'll wait with merging for a bit.

### DIFF
--- a/.github/workflows/cron.yml
+++ b/.github/workflows/cron.yml
@@ -1,0 +1,61 @@
+name: Cron
+
+on:
+  schedule:
+    # every friday at 10:00 am
+    - cron: '0 10 * * 5'
+
+jobs:
+  check:
+    name: Check
+    runs-on: ubuntu-latest
+    strategy:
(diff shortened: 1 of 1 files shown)


.github/workflows/cron.yml
```diff
@@ -0,0 +1,61 @@
...
+    - cron: '0 10 * * 5'
+
+jobs:
+  check:
+    name: Check
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        rust:
+          - 1.46.0
```
szarykott, line 15: I think it is below MSRV from README, was it intentional?

matthiasbeyer: I guess the MSRV was updated after this PR was created. I will update this PR. Thanks for pinging me!

PR #292, opened 2022-02-24
add prefix separator

### LINKED ISSUE #291: Regression in nested environment paths
In config 0.11, our app used an environment source

```rust
Environment::with_prefix("app").separator("__")
```

Which would match environment vars like `APP_TLS__CERTIFICATE` (into `tls > certificate`).

This no longer works due to the group separator using the specified separator, as opposed to always being `_` https://github.com/mehcode/config-rs/blob/30d87a311929f5382cb4ecd1de62f20d2dddfb86/src/env.rs#L123

This forces me to update my environments to `APP__TLS__CERTIFICATE` which is not the end of the world but it slows adoption of the new version.

I'm not sure on the best solution, but maybe there could be a new `prefix_separator` field. This can be set automatically by the `separator` method too so that it doesn't break any existing 0.12 apps

conradludgate:
Should fix #291

### DIFF
--- a/src/env.rs
+++ b/src/env.rs
@@ -16,6 +16,9 @@ pub struct Environment {
     /// For example, the key `CONFIG_DEBUG` would become `DEBUG` with a prefix of `config`.
     prefix: Option<String>,
 
+    /// Optional character sequence that separates the prefix from the rest of the key
+    prefix_separator: Option<String>,
+
     /// Optional character sequence that separates each key segment in an environment key pattern.
     /// Consider a nested configuration such as `redis.password`, a separator of `_` would allow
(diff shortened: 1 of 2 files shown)


matthiasbeyer (approved):
Awesome that you provided a patch right away!

Lets see what CI tells us.

conradludgate:
🤦 that clippy warning is annoying. My editor told me to do exactly the opposite 😅

matthiasbeyer (approved):

jrconlin:
Hi. Unfortunately, this bug impacts our code update. Is there any chance you could tag a new release with this fix?

PR #297, opened 2022-03-11
chore: update dependencies.

jaudiger:
This commit updates few dependencies to their latest version, and remove the Z version (in the X.Y.Z nomenclature) to always use the latest bug fix version available. And reduce for an application using config-rs to not build two times a dependency (I mean in two different version).

Plus, I remove a comment since the upgrade to Rust 1.49 here (https://github.com/mehcode/config-rs/commit/a7c6cc7daab224ff653d2b908ed9590ce3f6934b)

If I did something wrong, just let me know.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,30 +23,30 @@ json5 = ["json5_rs"]
 preserve_order = ["indexmap", "toml/preserve_order", "serde_json/preserve_order", "ron/indexmap"]
 
 [dependencies]
-async-trait = "0.1"
-lazy_static = "1.4"
-serde = "1.0"
-nom = "7"
+async-trait = "0.1.52"
+lazy_static = "1.4.0"
+serde = "1.0.136"
+nom = "7.1.0"
 
-toml = { version = "0.5", optional = true }
-serde_json = { version = "1.0", optional = true }
-yaml-rust = { version = "0.4", optional = true }
-rust-ini = { version = "0.18", optional = true }
-ron = { version = "0.7", optional = true }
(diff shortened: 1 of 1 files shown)


matthiasbeyer:
AFAIK, cargo automatically uses the latest version of a dependency if it is compatible (so for `foo = "1.0.0"`, a version `1.0.99` is compatible and would be used). So there would be no point in removing the patchlevel version part, right? :thinking:

jaudiger:
Gotcha, you are totally right.

So, should I revert the change relating to the bug fix version ? IMO since a bug fix is as its name states, just a bug fix, we could avoid to set the last version number here and just keep the two first ones. Since no breaking change should be introduced in a bug fix release. But I'll follow your opinion here !

jaudiger:
I just pushed a new commit to set the bug fix version to the latest available. I guess it's better to obtain reproducible build and tests when developing on the library.

matthiasbeyer:
Hm. To be honest I wouldn't do that. This might break user experience, if a user of this library needs, for whatever reason, to depend on some old version of one of our dependencies, they cannot use this library anymore because it pulls in a newer version.

I actually had such a problem [some time ago](https://github.com/science-computing/butido/blob/master/Cargo.toml#L94). This was due to licensing issues... the point is that it does not cost us anything to keep it as is, but it may cost a user some serious effort to work around things if we change this.

jaudiger:
Ok, no problem at all ! I can just close this PR. You're right, we should not change thing which are already working as expected.

matthiasbeyer:
Thanks for understanding! :heart: :partying_face:

Issue #302, opened 2022-03-18
Add set_override_optional combinator to ConfigBuilder


saraiva132:
Hey :)

Using ConfigBuilder and in case I have overrides that are optional, i.e. env_vars which may not be there, it would be quite idiomatic to add this combinator to avoid having:

` if let Ok(my_var) = b.set_override("region", region)?;`

In my case I would just convert my `Result` to `Option`

```
Config::builder()
            //Merge default config folder
            .add_source(default_config_file)
            .set_override_optional(my_optional_field)
```

I can submit a PR for this if you like :) Perhaps also for `default_optional`

WDYT?

matthiasbeyer:
Hey! Thanks for opening this issue. I am not sure what the actual implications of having an optional overwrite would be? Either you overwrite a value, or you don't. What would an optional overwrite actually be?

saraiva132:
@matthiasbeyer If optional value exists override otherwise do nothing.

matthiasbeyer:
And where would such an optional value come from? Do you have an example? :thinking:

saraiva132:
My use case is as mentioned.

```
let a = env::var("MY_INJECTED_VAR_A").ok();
let b = env::var("MY_INJECTED_VAR_B").ok();
let c = env::var("MY_INJECTED_VAR_C").ok();
Config::builder()
            //Merge default config folder
            .add_source(default_source)
            .set_override_optional("a", a)?
            .set_override_optional("b", b)?
            .set_override_optional("c", c)?;
```

Avoids this ugly syntax

```
        let builder = match env::var("MY_INJECTED_VAR_A") {
            Ok(a) => builder.set_override("a", a)?,
            Err(_) => builder,
        }
```

Perhaps this use case is not enough. I admit I am not able to discern others.

matthiasbeyer:
Why would you not [read environment variables](https://docs.rs/config/latest/config/struct.Environment.html) via `config`?

saraiva132:
I also read them via `config` but unfortunately a couple of them do not follow any pattern and must be individually loaded. `Environment` in config is not "powerful" enough for that

zhrebicek:
I guess it makes sense to have something like this.
Good example is some multi language environment having lot of shared environment variables where you do not want to add arbitrary prefix like `APP_` but also do not want to load them all blindly.

matthiasbeyer:
Hm. I think I would like to see a PR for this, yes.

PR #303, opened 2022-03-18
302-add-set-override-option-combinator

saraiva132:
Based on discussion in  #302. 

@matthiasbeyer I checked `Environment` and I do not believe it is not designed to support loading single env vars. This is just an example of how the method could look like.

### DIFF
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -173,6 +173,26 @@ impl<St: BuilderState> ConfigBuilder<St> {
             .insert(Expression::from_str(key.as_ref())?, value.into());
         Ok(self)
     }
+
+    /// Sets an override if value is Some(_)
+    ///
+    /// This function sets an overwrite value if Some(_) is passed. If None is passed, this function does nothing.
+    /// It will not be altered by any default, [`Source`] nor [`AsyncSource`]
+    ///
+    /// # Errors
+    ///
+    /// Fails if `Expression::from_str(key)` fails.
+    pub fn set_override_option<S, T>(mut self, key: S, value: Option<T>) -> Result<Self>
+    where
+        S: AsRef<str>,
+        T: Into<Value>,
+    {
+        if let Some(value) = value {
+            self.overrides
+                .insert(Expression::from_str(key.as_ref())?, value.into());
+        }
+        Ok(self)
+    }
 }
 
 impl ConfigBuilder<DefaultState> {


src/builder.rs
```diff
@@ -173,6 +173,26 @@ impl<St: BuilderState> ConfigBuilder<St> {
             .insert(Expression::from_str(key.as_ref())?, value.into());
         Ok(self)
     }
+
+    /// Sets an override if value is set
```
matthiasbeyer, line 177: I would like you to rephrase this into `Sets an override if value is Some(_)`. please. :smile:

src/builder.rs
```diff
@@ -173,6 +173,26 @@ impl<St: BuilderState> ConfigBuilder<St> {
...
+    ///
+    /// This function sets an overwrite value if value is not None, otherwise it does nothing.
```
matthiasbeyer, line 179: Also, don't use the negative case here. Better: "..sets an overwrite value if Some(_) is passed. If None is passed, this function does nothing".

matthiasbeyer (changes requested):
I think the change is simple enough to be merged, safe the two comments I had. Besides that :+1:

src/builder.rs
```diff
@@ -173,6 +173,26 @@ impl<St: BuilderState> ConfigBuilder<St> {
             .insert(Expression::from_str(key.as_ref())?, value.into());
         Ok(self)
     }
+
+    ///Sets an override if value is Some(_)
```
matthiasbeyer, line 177: Missing whitespace after `///`.

matthiasbeyer (changes requested):
One little nitpick left, I would really appreciate if you could fix this :smile: :pray:

matthiasbeyer:
Besides the one comment this looks good, feel free to squash (including the change for the comment from above)!

src/builder.rs
saraiva132: No worries. Comment formatting is important. My bad I did not notice!

saraiva132:
I failed my squash miserably, can you squash for me on merge ? :)  🙇

matthiasbeyer:
Not on merge, because squash merging is bad. I rebased for you and pushed to your PR branch.

matthiasbeyer:
The CI failure comes from something else, which I fix in #304 - I wil rebase this PR for you after that other fix is merged!

matthiasbeyer (approved):

PR #306, opened 2022-03-27
fix: dot in config name

CfirTsabari:
Rust std `set_extension` is either add an extension if none exists, or
edit the current one if such exists.

This caused a mishandling when user is using a name with a dot,
part of the name was treated as an extension, and be overwritten by
the different format extensions.

So manually *adding* a new dummy extension will cause the current code,
to behave as expected, since it will always overwrite the new dummy
extension and not part of the name.

### DIFF
--- a/src/file/source/file.rs
+++ b/src/file/source/file.rs
@@ -27,7 +27,7 @@ impl FileSourceFile {
     where
         F: FileStoredFormat + Format + 'static,
     {
-        let mut filename = if self.name.is_absolute() {
+        let filename = if self.name.is_absolute() {
             self.name.clone()
         } else {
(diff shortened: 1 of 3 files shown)


matthiasbeyer:
Sorry but I do not understand why this change is necessary. Could you explain a bit more (prefferably in the commit message)? And maybe provide tests for both bad cases and good cases? :thinking:

CfirTsabari:
@matthiasbeyer hi,
changed to commit message.
regarding the tests I have added a case for the bad case but the good cases is all of the other existing tests.

matthiasbeyer (approved):

matthiasbeyer:
Thanks! I really like the commit message, thanks for improving it! :heart: :tada:

PR #307, opened 2022-03-30
Fixed type in ConfigBuilder docs

bicarlsen:
Fixed a simple type in the `ConfigBuilder` docs, `alaways` -> `always`


matthiasbeyer:
Nice catch, thanks! :+1:

PR #318, opened 2022-04-22
Add Value::origin()

### LINKED ISSUE #97: Expose value origin
It could be as simple as adding "pub", but it should probably be hidden in an accessor.

    -    origin: Option<String>,
    +    pub origin: Option<String>,

(this is what I do now, to move forward with what I'm doing)

I can see a case for making it a thing that needs to be enabled, (elided at run-time by default) or require the use of a different type, (elided at compile-time, by default) but not having _any_ way to figure out where a value comes from isn't working for me.

matthiasbeyer:
Closes #97

### DIFF
--- a/src/value.rs
+++ b/src/value.rs
@@ -210,6 +210,11 @@ impl Value {
         }
     }
 
+    /// Get the description of the original location of the value.
+    pub fn origin(&self) -> Option<&str> {
+        self.origin.as_ref().map(AsRef::as_ref)
+    }
+
     /// Attempt to deserialize this value into the requested type.
     pub fn try_deserialize<'de, T: Deserialize<'de>>(self) -> Result<T> {
         T::deserialize(self)


xuorig:
Hi @matthiasbeyer we're interested in using `origin()` in our code base, but it looks like it is not exposed in the latest release, even though `0.13.3` succeeds this PR. Any chance you would be willing to cut a new `0.13.x` release soon?

matthiasbeyer:
Yes, we had a bunch of changes that should be released I guess... at least the dependency updates, but also this change here.

I will publish 0.14.0 soonish (probably this evening, but feel free to ping me again if it didn't happen until sunday)! :laughing:

xuorig:
Hi @matthiasbeyer, gentle ping on this 🙇

matthiasbeyer:
Hah, well I guess sunday was a long time ago.
I hope we can get some stuff in (https://github.com/mehcode/config-rs/milestone/4) and I guess after that a release is really much overdue.

polarathene:
> I hope we can get some stuff in (https://github.com/mehcode/config-rs/milestone/4)

I should be addressing those by this weekend, just getting a good understanding with the MSRV + `Cargo.lock` / CI concerns being tackled elsewhere for `0.13.x` first. Once I can chime in with my input on that, I'll get back to my own contributions 👍

matthiasbeyer:
Awesome, thanks a lot for that!

PR #334, opened 2022-05-08
errors: clarify names of integer types

martinvonz:
I got this error message:

```
invalid type: integer 64 bit `5`, expected an array
```

It took me a while to figure out that it wasn't talking about bit
number 5 of the number 64, but the 64-bit number `5`. I think it will
be clearer like this:

```
invalid type: 64-bit integer `5`, expected an array
```

So that's what this patch implements.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -23,10 +23,10 @@ impl fmt::Display for Unexpected {
     fn fmt(&self, f: &mut fmt::Formatter) -> result::Result<(), fmt::Error> {
         match *self {
             Unexpected::Bool(b) => write!(f, "boolean `{}`", b),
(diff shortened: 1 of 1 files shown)


matthiasbeyer (approved):

matthiasbeyer:
Thank you for this PR and _especially_ thank you for taking the time to explain why you did this in the commit message. Highly appreciated!

matthiasbeyer:
The CI fail is unrelated, feel free to simply rebase this to latest master and it should be green then :laughing:

martinvonz:
> The CI fail is unrelated, feel free to simply rebase this to latest master and it should be green then 😆

That's now done.

matthiasbeyer (approved):

PR #342, opened 2022-05-31
Add test for integer as key in YAML

### LINKED ISSUE #341: Deserializing maps with integer keys fails silently
Attempting to deserialize a map with an integer key fails, but does not produce an error. Putting quotes around the key makes it work as expected. This is using config `0.13.1`.

```rust
#![allow(dead_code, unused)]
use std::collections::HashMap;

use config::Config;
use serde::Deserialize;

#[derive(Debug, Deserialize)]
struct Outer {
    inner_string: HashMap<String, Inner>,
    inner_int: HashMap<u32, Inner>,
}

#[derive(Debug, Deserialize)]
struct Inner {
    member: String,
}

fn main() {
    let config = Config::builder()
        .add_source(config::File::with_name("config.yaml"))
        .build()
        .unwrap();
    println!("{:?}", config.try_deserialize::<Outer>().unwrap());
}
```
```yaml
# config.yaml
inner_int:
  "1":
    member: "Test Int 1"
  2:
    member: "Test Int 2"
inner_string:
  stuff:
    member: "Test String"
```
Output:
```
Outer { 
  inner_string: {
    "stuff": Inner { 
      member: "Test String" }
    }, 
  inner_int: {
    1: Inner { 
      member: "Test Int 1" 
    }
  }
}
```

matthiasbeyer:
WIP to fix #341 

@emgrav can you please check whether I do things correctly in the test here? From what I see, parsing an integer as key in YAML works as expected with the `yaml_rust` crate, but strangely using the `"1"` key it does not work as expected. That's why I think I've missed something here.

### DIFF
--- a/tests/int_key.rs
+++ b/tests/int_key.rs
@@ -1,0 +1,96 @@
+#![allow(dead_code, unused)]
+use std::collections::HashMap;
+
+use serde::Deserialize;
+use yaml_rust::YamlLoader;
+use yaml_rust::Yaml;
+
+use config::Config;
+
+#[derive(Debug, Deserialize)]
+struct Outer {
+    inner_string: HashMap<String, Inner>,
+    inner_int: HashMap<u32, Inner>,
+}
+
+#[derive(Debug, Deserialize)]
+struct Inner {
+    member: String,
+}
+
+const CONFIG: &str = r#"
+inner_int:
+  "1":
+    member: "Test Int 1"
+  2:
+    member: "Test Int 2"
+inner_string:
+  stuff:
+    member: "Test String"
+"#;
+
+#[test]
+fn test_int_key() {
+    let config = Config::builder()
+        .add_source(config::File::from_str(CONFIG, config::FileFormat::Yaml))
+        .build()
+        .unwrap();
+
+    let outer: Outer = config.try_deserialize().unwrap();
+    assert_eq!(outer.inner_string.get("1").unwrap().member, "Test Int 1");
+    assert_eq!(outer.inner_int.get(&2).unwrap().member, "Test Int 2");
+}
+
+#[test]
+fn test_yaml_parsing_int_key() {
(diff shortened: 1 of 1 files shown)


tests/int_key.rs
```diff
@@ -0,0 +1,96 @@
...
+
+#[test]
+fn test_int_key() {
+    let config = Config::builder()
+        .add_source(config::File::from_str(CONFIG, config::FileFormat::Yaml))
+        .build()
+        .unwrap();
+
+    let outer: Outer = config.try_deserialize().unwrap();
+    assert_eq!(outer.inner_string.get("1").unwrap().member, "Test Int 1");
```
emgrav, line 40: I don't know whether the key `"1"` being parsed into an int is actually correct behaviour, I'm not very familiar with yaml. However, it seems this is the case currently.

Anyway, assuming that behaviour is wanted - this test fails because it gets from the wrong map, and when getting from the correct map it expects a `&u32` instead of a `&str`.

```suggestion
    assert_eq!(outer.inner_int.get(&1).unwrap().member, "Test Int 1");
```

tests/int_key.rs
```diff
@@ -0,0 +1,96 @@
...
+
+    let member = int_hash.get(&Yaml::String("member".to_string()));
+    assert!(member.is_some());
+    let member = member.unwrap();
+
+    assert_eq!(member, &Yaml::String("Test Int 2".to_string()));
+}
+
+#[test]
+fn test_yaml_parsing_string_key() {
```
emgrav, line 72: I'm not sure what this is meant to test, I think things have gotten a little mixed up here since there's no `"1"` key in `inner_string`.

emgrav (changes requested):
I've checked as well as I can, but I must admit I don't know yaml well enough to be certain about any of this nor about what the desired behavior is. I hope it's at least somewhat helpful.

matthiasbeyer:
Closing this as superceede3d by #343

Issue #93, opened 2019-01-27
Deserializing into a type with an unsigned integer doesn't give the expected result


repnop:
Minimal example:

`main.rs`
```rust
use config::Config;
use serde::Deserialize;

#[derive(Deserialize)]
struct Foo {
    bar: u64,
}

fn main() {
    let mut settings = Config::new();
    settings.merge(config::File::with_name("test")).unwrap();
    
    let foo: Foo = settings.try_into().unwrap();
    println!("{}", foo.bar);
}
```
`test.toml`
```toml
bar = -1
```

Running this code produces `18446744073709551615` instead of erroring that the value was outside of the range of `u64`

jcrevier:
Additionally, related to this is that values greater than `i64::max_value()` will fail to de-serialize into a u64.  The problem seems to be the Deserialize implementation uses the [`into_int()` ](https://docs.rs/config/0.10.1/config/struct.Value.html#method.into_int)method that converts everything into i64 then casts back to the needed type.

kesyog:
Updated example:

```rs
use config::Config;
use serde::Deserialize;

#[derive(Deserialize)]
struct Foo {
    bar: u64,
}

fn main() {
    let settings = Config::builder()
        .add_source(config::File::with_name("test"))
        .build()
        .unwrap();

    let foo: Foo = settings.try_deserialize().unwrap();
    println!("{}", foo.bar);
}
```

PR #353, opened 2022-06-28
Use TryInto for more permissive deserialization for integers

### LINKED ISSUE #352: Deserialization fails if set_default is passed an unsigned integer
Small motivating example that demonstrates the issue:

```rs
use config::{Config, Environment};
use serde::Deserialize;
use std::error::Error;

// Using any unsigned type here leads to a runtime deserialization error
// Runs fine when using i8, i16, i32, i64
const DEFAULT_BAR: u32 = 30;

fn main() -> Result<(), Box<dyn Error>> {
    #[derive(Deserialize, Debug)]
    struct Foo {
        bar: u32,
    }
    let config: Foo = Config::builder()
        .add_source(Environment::with_prefix("TEST"))
        .set_default("bar", DEFAULT_BAR)?
        .build()?
        .try_deserialize()?;
    dbg!(config.bar);
    Ok(())
}
```
Output:
```
Error: invalid type: unsigned integer 64 bit `30`, expected an signed 64 bit or less integer for key `bar`
```

While looking into potentially fixing it, I noticed that `Value::into_uint` is defined but never used. Modifying the `deserialize_u32` method to use `into_uint` instead of `into_int` fixes this issue, but breaks some existing tests. I'm guessing that's because of the next observation:

`Value::into_int` and `Value::into_uint` both seem overly strict. For example, calling `Value::into_uint()` when the kind is a `ValueKind::I64` _always_ fails. Shouldn't we try a conversion using the `TryInto` trait and only fail if the conversion fails?

kesyog:
* Attempt to convert between integer types using `TryInto`-based
conversions rather than blanket failing for some source and
destination types.
* Use `into_uint` instead of `into_int` in `Value` Deserialize
implementations for unsigned integer types. Previously, we were
converting from signed types to unsigned types using `as`, which can
lead to surprise integer values conversions (#93).

Fixes #352 and #93


kesyog:
TODO: Add some unit tests around #352 and #93

src/de.rs
```diff
@@ -62,25 +62,25 @@ impl<'de> de::Deserializer<'de> for Value {
     #[inline]
     fn deserialize_u8<V: de::Visitor<'de>>(self, visitor: V) -> Result<V::Value> {
         // FIXME: This should *fail* if the value does not fit in the requets integer type
```
kesyog, line 64: Now's not a bad time to take of these FIXME's by replacing these `as` casts with `try_into`. I could roll that into this PR if desired.

kesyog:
@matthiasbeyer, I've cleaned up my draft implementation and added a few tests. Your review and CI workflow approval would be appreciated.

matthiasbeyer (approved):
Thank you very much for working on this!

kesyog:
No problem! It looks like all checks have passed and it's ready for merge whenever you're ready. It'd be great to get the new tests in before #338 to ensure that the functionality doesn't regress.

matthiasbeyer:
Yes! #338 has to be rebased onto latest master now, of course :laughing:

Issue #321, opened 2022-04-23
Re-thinking config-rs


matthiasbeyer:
# [Project: Re-thinking config-rs](https://github.com/mehcode/config-rs/projects/1)

I want to rethink how this crate works, what this crate provides in means of features and what it does _purposely_ not provide.

This issue is a tracking issue for the whole [project](https://github.com/mehcode/config-rs/projects/1).

## Current features (as of 0.13.x)

* Deserialization from various formats
  * Toml
  * Json
  * INI
  * Yaml
  * Ron
  * Json5
* "Layering" of configuration
* Different kinds of config sources
  * sync/async
  * Environment
* Setting defaults
* Setting overwrites
* Accessing fields using JSONPath-style selectors

## Accepted future features

As soon as a feature suggestion is discussed and agreed upon with the config-rs community, it will be moved to the "Accepted" Panel in [the project board](https://github.com/mehcode/config-rs/projects/1) and listed here.

# Help wanted

Please **open one issue per idea** for features you want to see in the config-rs implementation of the future.
No idea is too small for an issue.
  
---

This issue will be constantly updated!

matthiasbeyer:
The scope of this library is loading (from different sources) configuration and making it available to the running application conveniently, for rust application code.
So yes, this is for runtime configuration loading and parsing.

conradludgate:
If it helps, I'll document how config is currently used in a couple of my tools:

### At work, where we deploy web applications in kubernetes deployments:

Our main configuration is built into a yaml file.
We also load some secrets from vault into some more yaml files. These are all injected into the container.
Lastly, there are some extra secrets that we encode in environment variables.

### For https://github.com/ellie/atuin (cli app and server):

For client, the intent is that the user can modify their `config.toml` file in order to change how atuin runs. For development purposes of quick testing, we also have the environment parsing enabled.

For the server, there's only a small set of configuration, but similar idea where we have both a `server.toml` and environment parsing. Depending on how a user wants to host their server, one might be easier to configure than the other.

---

So far, I've never made use of the async (read remote http) configuration. For my work production apps, we don't allow defaults (mis-configuration is a bug). But for atuin, almost everything has a default.

szarykott:
I'd like to add something to discussion, it is not directly connected to this crate, hence I am not opening a new issue.

[...]

PR #354, opened 2022-06-30
Fix uppercase lowercase isses

### LINKED ISSUE #340: Environment variable names are converted to lower case
Environment variable names are converted to lower case internally. This is surprising undocumented behavior. We want to be able to override values from a configuration file with environment variables. It would be natural to use the same spelling (not counting the prefix), but this doesn't work if the configuration file keys are upper case. To make it work, the configuration file keys must be lower case.

config.toml
```bash
FOO="You can't override me!"
bar="You can override me just fine."
```
In the shell
```bash
APP_FOO="Why are you ignoring me?" APP_BAR="Yay! New bar." ./target/debug/app
```
I was using TOML for the config file and bash for the shell, if it matters.

YounessBird:
fix #340

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,8 +23,10 @@
 [RON]: https://github.com/ron-rs/ron
 [JSON5]: https://github.com/callum-oakley/json5-rs
 
-Please note that this library can not be used to write changed configuration
-values back to the configuration file(s)!
+Please note this library
+
+ - can not be used to write changed configuration values back to the configuration file(s)!
+ - Is case insensitive and all the keys are converted to lowercase internally
 
 ## Usage
 
--- a/src/config.rs
+++ b/src/config.rs
@@ -110,7 +110,8 @@ impl Config {
     where
         T: Into<Value>,
     {
-        self.defaults.insert(key.parse()?, value.into());
+        self.defaults
+            .insert(key.to_lowercase().as_str().parse()?, value.into());
 
         #[allow(deprecated)]
         self.refresh()
@@ -129,7 +130,8 @@ impl Config {
     where
         T: Into<Value>,
     {
-        self.overrides.insert(key.parse()?, value.into());
+        self.overrides
+            .insert(key.to_lowercase().as_str().parse()?, value.into());
 
         #[allow(deprecated)]
         self.refresh()
@@ -137,7 +139,7 @@ impl Config {
 
     #[deprecated(since = "0.12.0", note = "please use 'ConfigBuilder' instead")]
     pub fn set_once(&mut self, key: &str, value: Value) -> Result<()> {
-        let expr: path::Expression = key.parse()?;
+        let expr: path::Expression = key.to_lowercase().as_str().parse()?;
 
         // Traverse the cache using the path to (possibly) retrieve a value
         if let Some(ref mut val) = expr.get_mut(&mut self.cache) {
@@ -149,6 +151,8 @@ impl Config {
     }
 
     pub fn get<'de, T: Deserialize<'de>>(&self, key: &str) -> Result<T> {
+        let k = key.to_lowercase();
+        let key = k.as_str();
         // Parse the key into a path expression
         let expr: path::Expression = key.parse()?;
 
--- a/src/de.rs
+++ b/src/de.rs
@@ -246,7 +246,7 @@ impl EnumAccess {
     fn variant_deserializer(&self, name: &str) -> Result<StrDeserializer> {
         self.variants
             .iter()
-            .find(|&&s| s == name)
+            .find(|&&s| s.to_lowercase() == name.to_lowercase()) // changing to lowercase will enable deserialization of lowercase values to enums
             .map(|&s| StrDeserializer(s))
             .ok_or_else(|| self.no_constructor_error(name))
     }
--- a/tests/Settings.json
+++ b/tests/Settings.json
@@ -15,5 +15,7 @@
       "username": "jsmith",
       "email": "jsmith@localhost"
     }
-  }
+  },
+  "FOO": "FOO should be overridden",
+  "bar": "I am bar"
 }
--- a/tests/Settings.json5
+++ b/tests/Settings.json5
@@ -16,5 +16,7 @@
       "username": "jsmith",
       "email": "jsmith@localhost",
     }
-  }
+  },
+  FOO: "FOO should be overridden",
+  bar: "I am bar",
 }
--- a/tests/Settings.yaml
+++ b/tests/Settings.yaml
@@ -12,3 +12,6 @@ place:
     name: John Smith
     username: jsmith
     email: jsmith@localhost
+# For override tests
+FOO: FOO should be overridden
+bar: I am bar
--- a/src/path/mod.rs
+++ b/src/path/mod.rs
@@ -120,7 +120,7 @@ impl Expression {
         match *self {
             Self::Identifier(ref id) => match root.kind {
                 ValueKind::Table(ref mut map) => Some(
-                    map.entry(id.clone())
+                    map.entry(id.to_lowercase())
                         .or_insert_with(|| Value::new(None, ValueKind::Nil)),
                 ),
 
@@ -131,7 +131,7 @@ impl Expression {
                 Some(value) => {
                     if let ValueKind::Table(ref mut map) = value.kind {
                         Some(
-                            map.entry(key.clone())
+                            map.entry(key.to_lowercase())
                                 .or_insert_with(|| Value::new(None, ValueKind::Nil)),
                         )
                     } else {
@@ -139,7 +139,7 @@ impl Expression {
 
                         if let ValueKind::Table(ref mut map) = value.kind {
                             Some(
-                                map.entry(key.clone())
+                                map.entry(key.to_lowercase())
                                     .or_insert_with(|| Value::new(None, ValueKind::Nil)),
                             )
                         } else {
@@ -194,7 +194,7 @@ impl Expression {
                     ValueKind::Table(ref incoming_map) => {
                         // Pull out another table
                         let target = if let ValueKind::Table(ref mut map) = root.kind {
-                            map.entry(id.clone())
+                            map.entry(id.to_lowercase())
                                 .or_insert_with(|| Map::<String, Value>::new().into())
                         } else {
                             unreachable!();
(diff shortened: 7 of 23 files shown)


YounessBird:
Todo:
 json, json5, ini, ron

pictographer (approved):
Looks good.

One suggestion. Mention that keys are case insensitive in the docs.

YounessBird:
Thanks. I will add your suggestion to the the chore list :)

README.md
```diff
@@ -9,22 +9,24 @@
...
+-   Read from environment
+-   Loosely typed — Configuration values may be read in any supported type, as long as there exists a reasonable conversion
+-   Access nested fields using a formatted path — Uses a subset of JSONPath; currently supports the child ( `redis.port` ) and subscript operators ( `databases[0].name` )
+
+[json]: https://github.com/serde-rs/json
+[toml]: https://github.com/toml-lang/toml
+[yaml]: https://github.com/chyh1990/yaml-rust
+[ini]: https://github.com/zonyitoo/rust-ini
+[ron]: https://github.com/ron-rs/ron
+[json5]: https://github.com/callum-oakley/json5-rs
```
matthiasbeyer, line 24: Why was this changed?

README.md
```diff
@@ -9,22 +9,24 @@
...
+
+Please note this library
+
+-   Can not be used to write changed configuration values back to the configuration file(s)!
+-   It's case insensitive and all the keys are converted to lowercase internally.
```
matthiasbeyer, line 29:
```suggestion
-   Is case insensitive and all the keys are converted to lowercase internally.
```

README.md
```diff
@@ -35,12 +37,12 @@ config = "0.13.1"
...
- - `yaml` - Adds support for reading YAML files
- - `toml` - Adds support for reading TOML files
- - `ron` - Adds support for reading RON files
- - `json5` - Adds support for reading JSON5 files
+-   `ini` - Adds support for reading INI files
+-   `json` - Adds support for reading JSON files
+-   `yaml` - Adds support for reading YAML files
+-   `toml` - Adds support for reading TOML files
+-   `ron` - Adds support for reading RON files
+-   `json5` - Adds support for reading JSON5 files
```
matthiasbeyer, line 45: This is also an unrelated change.

src/config.rs
```diff
@@ -149,6 +151,8 @@ impl Config {
     }
 
     pub fn get<'de, T: Deserialize<'de>>(&self, key: &str) -> Result<T> {
+        let k = key.to_lowercase();
+        let key = k.as_str();
```
matthiasbeyer, line 155:
```suggestion
        let key = key.to_lowercase().as_str();
```

Does this work?

src/de.rs
```diff
@@ -345,19 +345,8 @@ impl<'de> de::Deserializer<'de> for Config {
...
-            ValueKind::U64(i) => visitor.visit_u64(i),
-            ValueKind::U128(i) => visitor.visit_u128(i),
-            ValueKind::Boolean(b) => visitor.visit_bool(b),
-            ValueKind::Float(f) => visitor.visit_f64(f),
-            ValueKind::String(s) => visitor.visit_string(s),
-            ValueKind::Array(values) => visitor.visit_seq(SeqAccess::new(values)),
-            ValueKind::Table(map) => visitor.visit_map(MapAccess::new(map)),
-        }
+        // Delegate deserialization to Value
+        de::Deserializer::deserialize_any(self.cache, visitor)
```
matthiasbeyer, line 349: How is this change related to the uppercase-lowercase issue?

src/lib.rs
```diff
@@ -14,7 +14,7 @@
 //!  - Live watching and re-reading of configuration files
 //!  - Deep access into the merged configuration via a path syntax
 //!  - Deserialization via `serde` of the configuration or any subset defined via a path
-//!
+//! Please note that this crate is case insensitive and all the keys are converted to lowercase internally.
```
matthiasbeyer, line 17: I guess you can leave the blank line after the listing. AFAIK having a blank line after a listing is considered good style in markdown.

Care to change this? :laughing:

tests/Settings.json
```diff
@@ -1,19 +1,21 @@
...
+        "latitude": 10.3970522,
+        "favorite": false,
+        "reviews": 3866,
+        "rating": 4.5,
+        "creator": {
+            "name": "John Smith",
+            "username": "jsmith",
+            "email": "jsmith@localhost"
+        }
+    },
```
matthiasbeyer, line 18: AFAICS you only changed the indentation here. Care to revert this so we get a smaller diff?

tests/Settings.json5
```diff
@@ -1,20 +1,22 @@
...
+        latitude: 10.3970522,
+        favorite: false,
+        reviews: 3866,
+        rating: 4.5,
+        creator: {
+            name: "John Smith",
+            username: "jsmith",
+            email: "jsmith@localhost",
+        },
+    },
```
matthiasbeyer, line 19: Same as above.

tests/Settings.yaml
```diff
@@ -2,13 +2,16 @@ debug: true
...
+    name: Torre di Pisa
+    longitude: 43.7224985
+    latitude: 10.3970522
+    favorite: false
+    reviews: 3866
+    rating: 4.5
+    creator:
+        name: John Smith
+        username: jsmith
+        email: jsmith@localhost
```
matthiasbeyer, line 14: ...and here.

matthiasbeyer (changes requested):
So first of all, thanks for working on this. This is important to get right.

I have some remarks, nothing too critical.

Still, I am not sure whether this is the right way to go. Converting all things to lowercase seems ... I don't know to be honest.

I am willing to merge this (modulo some things that have to be cleared), if we do not find another way... that said, when (if) we re-do the crate (#321) we should get it right from the get-go and support case sensitivity as well as insensitivity (let the user decide). But that's of course for another day, I guess.

README.md
YounessBird: Unintentional change as I was trying to find the best place to add that the crate is case insensitive.

src/config.rs
YounessBird: No I got the  "temporary value dropped while borrowed (rust)" error.  If you know of a more elegant way I will be happy to amend it.

src/de.rs
YounessBird: Unrelated to the uppercase-lowercase issue. However, I felt it's an overkill to open an issue to point out this minor change, so while i was working on the same file I thought of introducing this change. If you feel like it's unnecessary or you prefer the old way I will be happy to remove it :) .

tests/Settings.json
YounessBird: there two more line that I have added to this file 
` "FOO": "FOO should be overridden",
    "bar": "I am bar"
 `
the extra space is introduced by running rustfmt

tests/Settings.json5
YounessBird: there two more line that I have added to this file 
` "FOO": "FOO should be overridden",
    "bar": "I am bar"
 `
the extra space is introduced by running rustfmt

tests/Settings.yaml
YounessBird: there two more line that I have added to this file 
` "FOO": "FOO should be overridden",
    "bar": "I am bar"
 `
the extra space is introduced by running rustfmt

README.md
matthiasbeyer: No I think you changed some whitespace here. For example in c767d28 and in 0a2e8eb I can see whitespace change.

Sorry for being so persistent on this, I just want to be sure that the diff only touches what needs to be touched, because a larger diff makes harder to review. Right now it is not that much, but I hope you see the point: If a PR runs for weeks and the diff gets more and more overhead, this results in a discouraging amount of review work :laughing: 

If you don't want to continue because of my persistence, I can fully understand this and, if you're okay, I can take over this PR!

src/de.rs
matthiasbeyer: >  you feel like it's unnecessary or you prefer the old way I will be happy to remove it :) .

If you don't mind, feel free to open a seperate PR for this and remove it from this PR.

If not, I can do this myself if you like! :laughing:

src/lib.rs
matthiasbeyer: Oh okay, interesting. Can you link me to the CI failure for this?

tests/Settings.json
matthiasbeyer: rustfmt changes a JSON file?

README.md
YounessBird: > No I think you changed some whitespace here. For example in https://github.com/mehcode/config-rs/commit/c767d285d694f43539e4e98b6580024644cab22a and in https://github.com/mehcode/config-rs/commit/0a2e8ebcfe3951ea33ab99a3782f9193dd87d9d3 I can see whitespace change.

Yes I see what you mean. I think that has to do with the extension I am using when I hit save, my bad.

> Sorry for being so persistent on this, I just want to be sure that the diff only touches what needs to be touched, because a larger diff makes harder to review. Right now it is not that much, but I hope you see the point...

You are absolutely right, and at the time when I pushed this I just didn't notice these extra unnecessary changes in my commit. Thank you for pointing them out, I am gonna try and amend them later.

src/de.rs
YounessBird: Yes I will make sure to remove this change from this pr :).

tests/Settings.json
YounessBird: Sorry I think that has to do with the extension I am using when I hit save, my bad.

src/lib.rs
YounessBird: Yes, I think is this [one](https://github.com/mehcode/config-rs/actions/runs/2615343121)

matthiasbeyer: Ah. Yes, I see what gave you that impression: If you look closely, you'll see that on line 11 of the log you linked, the line is removed, but on line 12 it is added again.

This is usually an indicator of trailing whitespace that was removed! So there was a trailing whitespace that got removed, which shows as the line removed (line 11) and it looks like the very same line was added (line 12), but it is one without a trailing whitespace.
Unfortunately it is not clearly visible in the log.

I don't mind this, so I will mark this as resolved. :laughing:

matthiasbeyer (approved):
This was laying around way too long. I'm gonna merge this now, sorry for me taking so long!

utkarshgupta137:
There is still a bug (as of 0.13.3).
Consider this:
```
let conf: HashMap<String, String> = Config::builder()
    .add_source(File::with_name(name))
    .add_source(Environment::with_prefix(prefix))
    .build()?
    .try_deserialize()?;
```
File:
```
KEY = "value1"
```
Env:
```
PREFIX_KEY=value2
```

Then:
Current behavior: conf has 2 keys: `KEY` & `key` with values `value1` & `value2`
Expected behavior: conf has 1 key: `KEY` with value `value2`

Issue #357, opened 2022-07-18
Unable to deserialize unsigned integers


thrykol:
When implementing `config::ValueKind` for a custom type, the error `invalid type: unsigned integer 64 bit '128', expected an signed 64 bit or less integer for key 'inner.unsigned'` is thrown if the custom type contains a unsigned integer.  The following is a minimum example:

```
#[derive(Deserialize, Eq, PartialEq, Debug)]
struct Container<T> {
    inner: T,
}

#[derive(Deserialize, Eq, PartialEq, Debug)]
struct Unsigned {
    unsigned: u16,
}

impl Default for Unsigned {
    fn default() -> Self {
        Self { unsigned: 128 }
    }
}

impl From<Unsigned> for config::ValueKind {
    fn from(unsigned: Unsigned) -> Self {
        let mut properties = HashMap::default();
        properties.insert(
            "unsigned".to_string(),
            config::Value::from(unsigned.unsigned),
        );

        Self::Table(properties)
    }
}

assert_eq!(
    Container {
        inner: Unsigned::default()
    },
    config::Config::builder()
        .set_default("inner", Unsigned::default())
        .unwrap()
        .build()
        .unwrap()
        .try_deserialize::<Container<Unsigned>>()
        .unwrap()
);
```

matthiasbeyer:
Hi! Can you tell me where exactly that error gets thrown?

thrykol:
> Hi! Can you tell me where exactly that error gets thrown?

Unfortunately, no, I was unable to trace the origin down.

matthiasbeyer:
I mean in the above code.

thrykol:
> I mean in the above code.

Sorry... totally misunderstood you.  There error is returned from `.try_deserialize::<Container<Unsigned>>()`

matthiasbeyer:
I just added a testcase derived from your code in #359. Lets see whether I can reproduce your issue... but so far it looks like I can't!

matthiasbeyer:
Yep, I [cannot reproduce](So the test we came for [succeeds](https://github.com/mehcode/config-rs/runs/7518454547?check_suite_focus=true#step:4:582)) your issue... Maybe I've missed something, feel free to have a look at #359 and review whether it is equal to your code.

thrykol:
[...]

PR #359, opened 2022-07-26
Add test to deserialize unsigned int

matthiasbeyer:
Derived testcase from #357


matthiasbeyer:
So the test we came for [succeeds](https://github.com/mehcode/config-rs/runs/7518454547?check_suite_focus=true#step:4:582)

PR #361, opened 2022-08-02
Fix json5 test: Remove artifact character

matthiasbeyer:
I am not sure what happened that these tests suddenly fail, or where the
artifact comes from...

@up9cloud you implemented this two years ago... maybe you have at least an idea? :thinking:


matthiasbeyer:
Required for #360

matthiasbeyer:
Waiting for #362

PR #362, opened 2022-08-02
Run clippy only on MSRV

matthiasbeyer:
This patch reverts the matrix for the clippy job to only run clippy on
our MSRV instead of stable, beta and nightly.

The recurring issue with this was that lints started failing in PRs that
couldn't have been fixed in master yet, because master was never tested
with a clippy that new (in case of nightly).

1.56.1 is good enough for now.

---

Closes #361


matthiasbeyer:
#361 was included here now.

Issue #369, opened 2022-08-23
Overwrite a value from toml with env


tekjar:
Hi! I'm trying to overwrite a variable in toml with env variable. Due to restrictions w.r.t underscores, I had to alias a field with serde.

```rust
 config::{Config, ConfigError, File, Environment};
use serde::{Serialize,  Deserialize};

#[derive(Debug, Deserialize, Clone)]
pub struct Settings {
    pub name: String,
    pub v4: TlsConfig,
}

#[derive(Debug, Serialize, Deserialize, Clone)]
pub struct TlsConfig {
    pub ca_path: String,
    #[serde(alias = "certpath")]
    pub cert_path: String,
    // #[serde(alias = "keypath")]
    pub key_path: String,
}

pub fn main() -> Result<(), ConfigError> {
    std::env::set_var("V4_CERTPATH", "Hello World");

    let s = Config::builder()
        .add_source(File::with_name("default.toml"))
        .add_source(Environment::default().separator("_"))
        .build()?;

    dbg!(&s);
    let v: Settings = s.try_deserialize()?;
    dbg!(v);
    Ok(())
}
```

This is leading go this error

```
Error: duplicate field `cert_path`
```

How do I fix this?

matthiasbeyer:
Interesting! I filed a PR that adds (and hopefully fixes at some point) a test for this in #370

nstinus:
I think I found a reasonable workaround: use `"__"` (2 `'_'`) as separator. That way the nested struct are appropriately set.

matthiasbeyer:
Okay, I'll close this then. Thanks for reporting back!

polarathene:
> Thanks for reporting back!

That wasn't the original reporter btw.

PR #370, opened 2022-08-23
Add test for overwriting key from env

matthiasbeyer:
To address #369

### DIFF
--- a/tests/overwrite_key.rs
+++ b/tests/overwrite_key.rs
@@ -1,0 +1,40 @@
+use config::{Config, File, Environment};
+use serde::{Serialize,  Deserialize};
+
+const CONFIG: &str = r#"
+name = "foo"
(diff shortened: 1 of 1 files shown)


tekjar:
That's fast! :). Thanks @matthiasbeyer

matthiasbeyer:
Don't get too excited, this is just a test. I do not have a fix yet :laughing:

tekjar:
Heheh just noticed this. Still, thanks! :)

tekjar:
I would like to send a PR. Can you point me where overwrites happen in the code path?

matthiasbeyer:
Okay, so the fix would be to not alias the field.

That's unfortunately the only fix I see right now because of the way `config` does deserialization. There's no fix that would make the code at hand work as expected.

PR #373, opened 2022-08-31
Update notify

matthiasbeyer:
Includes #372 and a fixup for the changed interface of the `notify` crate.


0xpr03:
Just a heads up: You used a debounced event previously, this is moved to debouncer-notify-mini. Notify only emits raw events with v5.

PR #376, opened 2022-09-17
Rethink

matthiasbeyer:
This is my "rethink" efforts.

As this effort is done to overthink the architecture of the whole codebase, it removes the old code entirely and replaces it with my re-think work.

I hope interested parties will have a look.

---

Filed as draft, because **this will be rebased as I see fit**. So do not base any work on this except you want to add something to this PR!


FlixCoder:
Do you think it would be possible to generate sample configs including doc comments with the new approach? Or would it still need a separate macro to get the doc comments?

I am thinking about writing a macro for generating sample configs with documentation of the fields.

matthiasbeyer:
I could imagine that it would be possible.

milesj:
Is there any progress on this? I'm looking to move away from figment, and this crate is probably the best choice, but based on all the refactor talk, I'm hesitant to move to it.

matthiasbeyer:
> Is there any progress on this? 

There's progress here: https://github.com/matthiasbeyer/config-rs-ng - still slow, but at least there's progress!


> I'm looking to move away from figment, and this crate is probably the best choice, but based on all the refactor talk, I'm hesitant to move to it.

The current config-rs version is still usable :laughing: - _If_ we decide to move to the rewritten version, there will be a grace period and (if possible) compatible interfaces to make the move less complicated!

milesj:
Thanks for the update, I'll keep an eye on it 👍

PR #380, opened 2022-09-28
env: add a 'translate_key' field to ease dealing with kebab-case

da-x:
This allows usage of `kebab-case` attribute in serde, allowing to mapping unambiguously into a config value given a multiple character separator.

    let environment = Environment::default()
        .prefix("PREFIX")
        .translate_key(TranslationType::Kebab)
        .separator("__");

### DIFF
--- a/src/env.rs
+++ b/src/env.rs
@@ -5,6 +5,10 @@ use crate::map::Map;
 use crate::source::Source;
 use crate::value::{Value, ValueKind};
 
+/// An environment source collects a dictionary of environment variables values into a hierarchical
+/// config Value type. We have to be aware how the config tree is created from the environment
+/// dictionary, therefore we are mindful about prefixes for the environment keys, level separators,
+/// encoding form (kebab, snake case) etc.
 #[must_use]
 #[derive(Clone, Debug, Default)]
 pub struct Environment {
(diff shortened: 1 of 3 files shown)


src/env.rs
```diff
@@ -5,6 +5,10 @@ use crate::map::Map;
 use crate::source::Source;
 use crate::value::{Value, ValueKind};
 
+/// An environment source collects a dictionary of environment variables values into a hierarchical
+/// config Value type. We have to be aware how the config tree is created from the environment
+/// dictionary, therefore we are mindful about prefixes for the environment keys, level separators,
+/// encoding form (kebab, snake case) etc.
```
matthiasbeyer, line 11: Could you please split this to a seperate commit?

src/env.rs
```diff
@@ -77,6 +86,11 @@ pub struct Environment {
     source: Option<Map<String, String>>,
 }
 
+#[derive(Clone, Debug)]
+pub enum TranslationType {
+    Kebab, // Translate '_' to '-'
```
matthiasbeyer, line 91: Would it make sense to add more here?

src/env.rs
```diff
@@ -197,6 +221,13 @@ impl Source for Environment {
             if !separator.is_empty() {
                 key = key.replace(separator, ".");
             }
+            if let Some(translate_key) = translate_key {
+                match translate_key {
+                    TranslationType::Kebab => {
+                        key = key.replace("_", "-");
+                    },
+                }
+            }
```
matthiasbeyer, line 230:
```suggestion
            if let Some(TranslationType::Kebab) = translate_key {
                key = key.replace("_", "-");
            }
```

But if we add more types then this is not possible of course. :laughing:

matthiasbeyer (changes requested):
I really like that you added tests as well!

Please have a look at my annotations. I will approve CI right now though, so we can catch things early.

src/env.rs
da-x: We can go even further, depending on [`convert_case` crate](https://docs.rs/convert_case/latest/convert_case/enum.Case.html) and pass a `convert_case::Case` value here instead (eliminating `TranslationType`). If doing this, should we add a feature flag with optional dependency?

matthiasbeyer: That sounds good, IMO. Yes, a feature flag would be nice!

da-x:
@matthiasbeyer, all requested done in another PR to fix branch name mixup. Here: https://github.com/mehcode/config-rs/pull/381

PR #381, opened 2022-09-29
env: add a 'convert_case' field to ease dealing with kebab-case

da-x:
This allows usage of `kebab-case` attribute in serde, mapping
unambiguously into a config value given a multiple character separator.

This also add the `convert-case` feature.

For example:

    let environment = Environment::default()
        .prefix("PREFIX")
        .translate_key(Case::Kebab)
        .separator("__");

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -42,3 +42,6 @@ pub use crate::format::Format;
 pub use crate::map::Map;
 pub use crate::source::{AsyncSource, Source};
 pub use crate::value::{Value, ValueKind};
+
+// Re-export
+pub use convert_case::Case;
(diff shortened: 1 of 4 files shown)


src/lib.rs
```diff
@@ -42,3 +42,6 @@ pub use crate::format::Format;
 pub use crate::map::Map;
 pub use crate::source::{AsyncSource, Source};
 pub use crate::value::{Value, ValueKind};
+
+// Re-export
+pub use convert_case::Case;
```
matthiasbeyer, line 47: That should be feature-gated, right?

matthiasbeyer (changes requested):

PR #398, opened 2022-12-04
Backport 379: Add clone trait to builder state

matthiasbeyer:
(cherry picked from commit 02c80a5c5a68022bd2d9eee7a67acff02062abb0)


matthiasbeyer:
Grml. Because this is a library crate and we do not have a `Cargo.lock`, `scoped-tls` gets updated in this PR which makes clippy fail because the MSRV for that crate was updated.

Meh.

That's not how it should work.

PR #413, opened 2023-01-17
Attach key to type error generated from Config::get_<type>()

yuja:
This commit extracts low-level get_value() to preserve the origin as much as possible. get::<Value>() would run Value-to-Value deserialization, and the origin field would be lost there.

For scalar types, we could instead leverage get::<T>() as the deserializer implemented for Value type invokes Value::into_<type>() function:

    fn get_string(&self, key: &str) -> Result<String> { self.get(key) }


matthiasbeyer (changes requested):
Sorry for not answering sooner.

I think I like what I see. Could you please rebase your PR to latest master, because we had a toml-rs update merged and that error messages you're testing for might have changed.

yuja:
Sure, rebased. The tests passed locally.

matthiasbeyer (approved):

matthiasbeyer:
Thanks for contributing! :tada:

Issue #414, opened 2023-01-18
Option<Decimal> always None


telcy:
config.yml
```
foo: 100.0
bar: 200.0
```

```
use rust_decimal::Decimal;
use serde::Deserialize;

#[derive(Debug, Deserialize)]
pub struct Config {
    pub foo: Decimal,
    pub bar: Option<Decimal>,
}
```

bar should be Some(200.0) but is always None.

matthiasbeyer:
Hi! Thanks for your report. Unfortunately I cannot reproduce, see #415

telcy:
Thank you. Appreciate your help.
It is all working as intended and it turns out my slightly more complex case had a bug.

Keep up the good work.

PR #415, opened 2023-01-19
Add test with rust_decimal

matthiasbeyer:
To test issue #414

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -55,3 +55,4 @@ lazy_static = "1"
 notify = "^5.0.0"
 temp-env = "0.3.0"
 log = { version = "0.4", features = ["serde"] }
+rust_decimal = "1.27.0"
--- a/tests/yaml-decimal.rs
+++ b/tests/yaml-decimal.rs
@@ -1,0 +1,28 @@
+#![cfg(feature = "yaml")]
+
+use rust_decimal::Decimal;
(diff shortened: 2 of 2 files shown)


matthiasbeyer:
Why the hell does it not find `rust_decimal`? Locally this works perfectly nice and `rust_decimal` is in `1.27.0`? :eyes:

matthiasbeyer:
Issue was on users side, not gonna merge this to not unnecessarily bloat the dev-dependencies of the project.

telcy:
According to rust_decimal docs: "The current minimum compiler version is 1.60.0 which was released on 2022-04-07."

That's why compiling for 1.59.0 failed.

matthiasbeyer:
But wouldn't cargo still find the crate and try to compile it and then fail? From what I see, it fails to actually find the version!

telcy:
I installed locally 1.59.0 and it failed with same output.

Probably just an early exit because of the rust version defined for the crate?
https://github.com/paupino/rust-decimal/blob/bc8b53a7ca697945642ebe3a44f38d9cd6f0bd83/Cargo.toml#L19

https://doc.rust-lang.org/cargo/reference/manifest.html#the-rust-version-field

matthiasbeyer:
Hm, okay. I didn't know cargo can do that. Thanks for investigating this!

Issue #417, opened 2023-01-19
Incorrect deserialization with yaml feature and keys containing periods


morsley:
When using `yaml` as a config format, the deserialization incorrectly stops at periods in keys.  I do not observe this behavior when manually deserializing the string with `serde_yaml` or other deserializers (such as `ruamel.yaml` or `pyyaml` in Python).  As a result, I am getting errors about keys not being present in the mapped value even though they definitely exist.

MCVE:

```yml
192.168.1.1:
  an_arr:
    - 1
    - 2
    - 3
  another_value: 10
```

```rs
#[derive(Debug, serde::Deserialize)]
struct Conf {
  an_arr: Vec<u32>,
  another_value: u32,
}

fn main() {
  let settings = config::Config::builder()
    .add_source(config::File::from("config.yml"))
    .build()
    .unwrap();

  println!(
    "{:?}",
    settings
      .try_deserialize::<HashMap<String, Conf>>()
      .unwrap()
  );
}
```

Result:

```
The application panicked (crashed).
Message:  called `Result::unwrap()` on an `Err` value: missing field `another_value`
Location: src/main.rs:51

  ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ BACKTRACE ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
                                ⋮ 9 frames hidden ⋮                               
  10: core::result::Result<T,E>::unwrap::hdf9dbc5c9c26ec5d
      at /rustc/90743e7298aca107ddaa0c202a4d3604e29bfeb6/library/core/src/result.rs:1113
  11: ui::main::he2f965d0d6a736c9
      at /home/redacted/src/main.rs:49
        47 │         println!(
        48 │             "{:?}",
        49 >             settings
        50 │                 .try_deserialize::<HashMap<String, Conf>>()
        51 │                 .unwrap()
```

matthiasbeyer:
Hi! Thanks for your report. Would you mind filing a patch showcasing this? I would be especially interested in a comparison of `config-rs` and the plain use of the yaml crate in use by `config-rs` (`serde_yaml` IIRC).

Can all be in one testcase, in one file in `/tests`.

If you cannot spare the time for this, just ping me and I'll do it myself.

Halkcyon:
I expected `serde_yaml`, too, but it appears `config-rs` is [using `yaml-rust`](https://crates.io/crates/yaml-rust) @matthiasbeyer 

[...]

PR #418, opened 2023-01-19
repro for #417

Halkcyon:


matthiasbeyer:
Nice. Could you also add one with `yaml-rust`, so that we have all three cases?

I will then take these patches and continue based on them... not sure how, but I guess I will think of something.

Halkcyon:
@matthiasbeyer not sure I understand; the default case is `yaml-rust` with the second test case being for `serde_yaml`.

matthiasbeyer:
Yes, I meant add another case where you use _only_ `yaml-rust`, without `config-rs` in between. This way we can verify that we don't mess things up in the `config-rs` implementation!

Halkcyon:
@matthiasbeyer Done.  Surprising result, the new test passes

tests/file_yaml.rs
```diff
@@ -35,6 +37,35 @@ fn make() -> Config {
...
+fn test_keys_with_periods_deserialize() {
+    let c = make();
+
+    let s: Settings = c.try_deserialize().unwrap();
+
+    assert_eq!(s.ip_key, "a string value");
+}
+
 #[test]
 fn test_file() {
```
matthiasbeyer, line 70: This test fails because the files in `tests/Settings` are not updated for the new key ("192.168.1.1")!

tests/file_yaml.rs
```diff
@@ -35,6 +37,35 @@ fn make() -> Config {
...
+    let c = YamlLoader::load_from_str(map).unwrap().first().unwrap().clone();
+
+    assert_eq!(c["192.168.1.1"].as_str().unwrap(), "a string value");
+}
+
+#[test]
+fn test_keys_with_periods_deserialize() {
+    let c = make();
+
+    let s: Settings = c.try_deserialize().unwrap();
```
matthiasbeyer, line 64: This test fails here because the `tests/Settings` files are not updated.

matthiasbeyer (changes requested):
The changes itself look good.

CI fails because the files in `tests/Settings` are not updated to match the expected `Settings` structure!

The commit linter also fails. Please:

* Rewrite the commits in imperative mood ("Add stuff" instead of "Adding stuff"). Don't hesitate to include details in the message body!
* Ensure the length of the subject line does not exceed 50 characters
* Make sure to sign off your commits

In general: https://cbea.ms/git-commit/ :wink: 

---

Or ping me if you don't care anymore, then I will take over this PR! :wink:

tests/file_yaml.rs
Halkcyon: @matthiasbeyer It was updated, but due to #417 it fails.  This is the bug reproduced.

https://github.com/Halkcyon/config-rs/blob/cdb945f8a2b2fd7223d518eec874e28e38e491fa/tests/Settings.yaml#L18

tests/file_yaml.rs
Halkcyon: It is updated.  It fails due to #417

tests/file_yaml.rs
matthiasbeyer: Ah, I guess I was not fully awake when reviewing this. Sorry about that.

matthiasbeyer:
So if I understand correctly: 

* Plain `yaml-rust` succeeds
* `config-rs` (which uses `yaml-rust`) fails

Okay, so we have to investigate why that is.

polarathene:
This is not specific to YAML parsing, but feature of `config-rs` with map lookup keys using dot notation.

---

> Okay, so we have to investigate why that is.

Failure is due to `.` not being a permitted raw identifier char (_accepting `.` here will fix this_):

https://github.com/mehcode/config-rs/blob/6946069755e4bc75af9b7ca678da66c055a0af16/src/path/parser.rs#L15-L25

Instead `.` is used to designate a child field (_accepting a `.` above will cause the test for this feature to fail_):

https://github.com/mehcode/config-rs/blob/6946069755e4bc75af9b7ca678da66c055a0af16/src/path/parser.rs#L38-L49

That above code is called from this method during config build:

https://github.com/mehcode/config-rs/blob/6946069755e4bc75af9b7ca678da66c055a0af16/src/source.rs#L30-L38

which multiple locations will call via this method:

https://github.com/mehcode/config-rs/blob/6946069755e4bc75af9b7ca678da66c055a0af16/src/source.rs#L20-L27

---

I'm not too familiar with `config-rs`, but vaguely recall some feature to access fields with dot notation. I assume it's for the value map as a string key lookup feature. Thus you probably can't have both here unless there was support to make the distinction 🤷‍♂️ 

I understand the intent was to deserialize with the serde rename feature to the desired field. This won't work AFAIK since the builder is creating this `cache` prior to deserializing which splits up the key:

```yml
# Example input
hello.world: example
192.168.1.1: a string value
```

<details>
<summary>Built config before deserializing to target config from user</summary>

```
A config: Config {
    defaults: {},
    overrides: {},
    sources: [],
    cache: Value {
        origin: None,
        kind: Table(
            {
                "hello": Value {
                    origin: None,
                    kind: Table(
                        {
                            "world": Value {
                                origin: Some(
                                    "hello.yaml",
                                ),
                                kind: String(
                                    "example",
                                ),
                            },
                        },
                    ),
                },
                "192": Value {
                    origin: None,
                    kind: Table(
                        {
                            "168": Value {
                                origin: None,
                                kind: Table(
                                    {
                                        "1": Value {
                                            origin: None,
                                            kind: Table(
                                                {
                                                    "1": Value {
                                                        origin: Some(
                                                            "hello.yaml",
                                                        ),
                                                        kind: String(
                                                            "a string value",
                                                        ),
                                                    },
                                                },
                                            ),
                                        },
                                    },
                                ),
                            },
                        },
                    ),
                },
            },
        ),
    },
}
```

</details>

<details>
<summary>Config format deserializes correctly before handed to builder as source input</summary>

```
Value {
    origin: Some(
        "hello.yaml",
    ),
    kind: Table(
        {
            "hello.world": Value {
                origin: Some(
                    "hello.yaml",
                ),
                kind: String(
                    "example",
                ),
            },
            "192.168.1.1": Value {
                origin: Some(
                    "hello.yaml",
                ),
                kind: String(
                    "a string value",
                ),
            },
        },
    ),
}
```

</details>

Issue #245, opened 2021-10-27
[feature request]: Support Gura


polarathene:
Would there be interest in adding support for [Gura](https://gura.netlify.app/)? It's similar to TOML and YAML, the link details where Gura addresses these two formats drawbacks. It seems like it'd be a good format to support?

It is actively maintained and there is [an official serde crate which should help?](https://github.com/gura-conf/serde-gura)

epage:
Note about adding a new format: traditionally, we've enabled all of them by default but #621 would have us re-evaluate that assumption.

PR #446, opened 2023-07-29
Added Gura support

mkvolkov:
Added support for Gura. Closes issue #245

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,7 +15,7 @@ edition = "2018"
 maintenance = { status = "actively-developed" }
 
 [features]
-default = ["toml", "json", "yaml", "ini", "ron", "json5", "ura", "convert-case", "async"]
+default = ["toml", "json", "yaml", "ini", "ron", "json5", "gura", "convert-case", "async"]
 json = ["serde_json"]
 yaml = ["yaml-rust"]
 ini = ["rust-ini"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,11 +15,12 @@ edition = "2018"
 maintenance = { status = "actively-developed" }
 
 [features]
-default = ["toml", "json", "yaml", "ini", "ron", "json5", "convert-case", "async"]
+default = ["toml", "json", "yaml", "ini", "ron", "json5", "ura", "convert-case", "async"]
 json = ["serde_json"]
 yaml = ["yaml-rust"]
 ini = ["rust-ini"]
 json5 = ["json5_rs"]
+ura = ["gura"]
 convert-case = ["convert_case"]
 preserve_order = ["indexmap", "toml/preserve_order", "serde_json/preserve_order", "ron/indexmap"]
 async = ["async-trait"]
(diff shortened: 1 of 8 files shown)


Cargo.toml
```diff
@@ -15,11 +15,12 @@ edition = "2018"
 maintenance = { status = "actively-developed" }
 
 [features]
-default = ["toml", "json", "yaml", "ini", "ron", "json5", "convert-case", "async"]
+default = ["toml", "json", "yaml", "ini", "ron", "json5", "ura", "convert-case", "async"]
```
matthiasbeyer, line 18: Why is the feature named "ura" instead of "gura"?

matthiasbeyer (changes requested):
Hi! Thanks for continuing!

I am off the grid for approx. The next week, so please expect delays in response.

One question (see review). There might be more, I cannot do in depth reviews now as I am on a train. Either way, thanks again for contributing!

src/file/format/mod.rs
```diff
@@ -27,6 +27,9 @@ mod ron;
 #[cfg(feature = "json5")]
 mod json5;
 
+#[cfg(feature = "ura")]
```
matthiasbeyer, line 30: It's also here.

src/file/format/mod.rs
```diff
@@ -55,6 +58,10 @@ pub enum FileFormat {
     /// JSON5 (parsed with json5)
     #[cfg(feature = "json5")]
     Json5,
+
+    /// GURA (parsed with ura)
+    #[cfg(feature = "ura")]
```
matthiasbeyer, line 63: And here.

src/file/format/mod.rs
```diff
@@ -81,6 +88,9 @@ lazy_static! {
         #[cfg(feature = "json5")]
         formats.insert(FileFormat::Json5, vec!["json5"]);
 
+        #[cfg(feature = "ura")]
+        formats.insert(FileFormat::Gura, vec!["ura"]);
```
matthiasbeyer, line 92: And here

src/file/format/mod.rs
```diff
@@ -117,6 +127,9 @@ impl FileFormat {
             #[cfg(feature = "json5")]
             FileFormat::Json5 => json5::parse(uri, text),
 
+            #[cfg(feature = "ura")]
```
matthiasbeyer, line 130: And here

matthiasbeyer (changes requested):

src/file/format/mod.rs
polarathene: If I'm not mistaken, this section is fine as `ura` as [`.ura` is the filename extension](https://gura.netlify.app/docs/conventions#filename-extension).

Friendly ping as changes were applied since and it'd be good to have the effort merged before it becomes too stale.

polarathene:
@matthiasbeyer friendly ping as this PR is getting stale (_already conflicts raised_).

Only one `ura` remains from your feedback, but it's [intentional as it's the file format extension](https://github.com/mehcode/config-rs/pull/446#discussion_r1313021487).

matthiasbeyer:
Yes this needs a rebase for resolving conflicts.

polarathene:
> Yes this needs a rebase for resolving conflicts.

Done https://github.com/mehcode/config-rs/pull/467

@mkvolkov can either update their branch here, otherwise my alternative one should be sufficient if they're unresponsive.

Cargo.toml
```diff
@@ -36,6 +37,7 @@ yaml-rust = { version = "0.4", optional = true }
 rust-ini = { version = "0.19", optional = true }
 ron = { version = "0.8", optional = true }
 json5_rs = { version = "0.4", optional = true, package = "json5" }
+gura = { version = "0.3.0", optional = true, package = "gura" }
```
polarathene, line 40:
```suggestion
gura = { version = "0.5", optional = true, package = "gura" }
```

PR #457, opened 2023-09-19
Support jsonc format

up9cloud:
I did a new commit, see https://github.com/mehcode/config-rs/pull/452

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -126,8 +126,9 @@ json = ["serde_json"]
 yaml = ["yaml-rust2"]
 ini = ["rust-ini"]
 json5 = ["json5_rs", "dep:serde-untagged"]
+jsonc = ["dep:jsonc-parser"]
 convert-case = ["convert_case"]
-preserve_order = ["indexmap", "toml?/preserve_order", "serde_json?/preserve_order", "ron?/indexmap"]
+preserve_order = ["indexmap", "toml?/preserve_order", "serde_json?/preserve_order", "jsonc-parser?/preserve_order", "ron?/indexmap"]
 async = ["async-trait"]
 toml = ["dep:toml"]
 
@@ -141,6 +142,7 @@ yaml-rust2 = { version = "0.10", optional = true }
 rust-ini = { version = "0.21", optional = true }
 ron = { version = "0.8", optional = true }
 json5_rs = { version = "0.4", optional = true, package = "json5" }
+jsonc-parser = { version = "0.26.3", optional = true }
 indexmap = { version = "2.10.0", features = ["serde"], optional = true }
 convert_case = { version = "0.6", optional = true }
 pathdiff = "0.2"
--- a/src/file/format/jsonc.rs
+++ b/src/file/format/jsonc.rs
@@ -1,0 +1,49 @@
+use std::error::Error;
+
+use crate::format;
+use crate::map::Map;
+use crate::value::{Value, ValueKind};
+use jsonc_parser::JsonValue;
+
+pub(crate) fn parse(
+    uri: Option<&String>,
+    text: &str,
+) -> Result<Map<String, Value>, Box<dyn Error + Send + Sync>> {
+    let value = match jsonc_parser::parse_to_value(text, &Default::default())? {
+        Some(json_value) => from_jsonc_value(uri, json_value),
+        None => Value::new(uri, ValueKind::Nil),
+    };
+    format::extract_root_table(uri, value)
+}
+
+fn from_jsonc_value(uri: Option<&String>, value: JsonValue<'_>) -> Value {
+    let vk = match value {
+        JsonValue::Null => ValueKind::Nil,
+        JsonValue::String(v) => ValueKind::String(v.to_string()),
+        JsonValue::Number(number) => {
+            if let Ok(v) = number.parse::<i64>() {
+                ValueKind::I64(v)
+            } else if let Ok(v) = number.parse::<f64>() {
+                ValueKind::Float(v)
+            } else {
+                unreachable!();
+            }
+        }
+        JsonValue::Boolean(v) => ValueKind::Boolean(v),
+        JsonValue::Object(table) => {
+            let m = table
+                .into_iter()
+                .map(|(k, v)| (k, from_jsonc_value(uri, v)))
+                .collect();
+            ValueKind::Table(m)
+        }
+        JsonValue::Array(array) => {
+            let l = array
+                .into_iter()
+                .map(|v| from_jsonc_value(uri, v))
+                .collect();
+            ValueKind::Array(l)
+        }
+    };
+    Value::new(uri, vk)
+}
--- a/src/file/format/mod.rs
+++ b/src/file/format/mod.rs
@@ -23,6 +23,9 @@ mod ron;
 #[cfg(feature = "json5")]
 mod json5;
 
+#[cfg(feature = "jsonc")]
+mod jsonc;
+
 /// File formats provided by the library.
 ///
 /// Although it is possible to define custom formats using [`Format`] trait it is recommended to use `FileFormat` if possible.
@@ -52,6 +55,10 @@ pub enum FileFormat {
     /// JSON5 (parsed with json5)
     #[cfg(feature = "json5")]
     Json5,
+
+    /// JSONC (parsed with jsonc)
+    #[cfg(feature = "jsonc")]
+    Jsonc,
 }
 
 pub(crate) fn all_extensions() -> &'static HashMap<FileFormat, Vec<&'static str>> {
@@ -78,6 +85,9 @@ pub(crate) fn all_extensions() -> &'static HashMap<FileFormat, Vec<&'static str>
 
         #[cfg(feature = "json5")]
         formats.insert(FileFormat::Json5, vec!["json5"]);
+
+        #[cfg(feature = "jsonc")]
+        formats.insert(FileFormat::Jsonc, vec!["jsonc"]);
 
         formats
     })
@@ -115,6 +125,9 @@ impl FileFormat {
             #[cfg(feature = "json5")]
             FileFormat::Json5 => json5::parse(uri, text),
 
+            #[cfg(feature = "jsonc")]
+            FileFormat::Jsonc => jsonc::parse(uri, text),
+
             #[cfg(all(
                 not(feature = "toml"),
                 not(feature = "json"),
@@ -122,6 +135,7 @@ impl FileFormat {
                 not(feature = "ini"),
                 not(feature = "ron"),
                 not(feature = "json5"),
+                not(feature = "jsonc"),
             ))]
             _ => unreachable!("No features are enabled, this library won't work without features"),
         }
--- a/tests/testsuite/file_jsonc.rs
+++ b/tests/testsuite/file_jsonc.rs
@@ -1,0 +1,205 @@
+#![cfg(feature = "jsonc")]
+
+use serde_derive::Deserialize;
+
+use config::{Config, File, FileFormat, Map, Value};
+use float_cmp::ApproxEqUlps;
+use snapbox::{assert_data_eq, str};
+
+/// Returns the path to a test config file with an optional suffix.
+///
+/// # Example
+/// If the current file path is `/workspace/config-rs/tests/testsuite/file_jsonc.rs`:
+/// ```
+/// let path = get_config_file_path("");
+/// assert_eq!(path, "/workspace/config-rs/tests/testsuite/file_jsonc.jsonc");
+///
+/// let path = get_config_file_path(".extra");
+/// assert_eq!(path, "/workspace/config-rs/tests/testsuite/file_jsonc.extra.jsonc");
+/// ```
+fn get_config_file_path(suffix: &str) -> String {
+    let path = std::path::Path::new(file!());
+    format!(
+        "{}/{}/{}{}.jsonc",
+        env!("CARGO_MANIFEST_DIR"),
+        path.parent().unwrap().to_str().unwrap(),
+        path.file_stem().unwrap().to_str().unwrap(),
+        suffix
+    )
+}
+
+#[test]
+fn test_file() {
+    #[derive(Debug, Deserialize)]
+    struct Place {
+        name: String,
+        longitude: f64,
+        latitude: f64,
+        favorite: bool,
+        telephone: Option<String>,
+        reviews: u64,
+        creator: Map<String, Value>,
+        rating: Option<f32>,
+    }
+
+    #[derive(Debug, Deserialize)]
+    struct Settings {
+        debug: f64,
+        production: Option<String>,
+        place: Place,
+        #[serde(rename = "arr")]
+        elements: Vec<String>,
+    }
+
+    let c = Config::builder()
+        .add_source(File::new(&get_config_file_path(""), FileFormat::Jsonc))
+        .build()
+        .unwrap();
+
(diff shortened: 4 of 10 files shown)


tests/file_jsonc.rs
```diff
@@ -52,7 +52,7 @@ fn test_file() {
     assert_eq!(s.place.telephone, None);
     assert_eq!(s.elements.len(), 10);
     assert_eq!(s.elements[3], "4".to_string());
-    if cfg!(feature = "preserve_order") {
+    if cfg!(feature = "TODO: preserve_order") {
```
matthiasbeyer, line 55: Why this?

src/file/format/mod.rs
```diff
@@ -81,6 +88,12 @@ lazy_static! {
         #[cfg(feature = "json5")]
         formats.insert(FileFormat::Json5, vec!["json5"]);
 
+        #[cfg(all(feature = "jsonc", feature = "json"))]
+        formats.insert(FileFormat::Jsonc, vec!["jsonc"]);
+
+        #[cfg(all(feature = "jsonc", not(feature = "json")))]
+        formats.insert(FileFormat::Jsonc, vec!["jsonc", "json"]);
```
matthiasbeyer, line 95: Do I understand correctly that the `jsonc` parser also parses `json` even if that feature is disabled?

I think that's counter-intuitive and we should only parse `jsonc` with the `jsonc` parser.

matthiasbeyer (changes requested):
I have some remarks. Also, the two commits could be squashed into one commit,... as well as further fixups of course!

Either way thanks for your contribution!

Cargo.toml
```diff
@@ -15,11 +15,12 @@ edition = "2018"
 maintenance = { status = "actively-developed" }
 
 [features]
-default = ["toml", "json", "yaml", "ini", "ron", "json5", "convert-case", "async"]
+default = ["toml", "json", "yaml", "ini", "ron", "json5", "jsonc", "convert-case", "async"]
```
epage: Note about adding a new format: traditionally, we've enabled all of them by default but #621 would have us re-evaluate that assumption.

up9cloud: Got it.

up9cloud:
Well, I merged the upstream and fixed the test cases.
However, I believe forcing lowercase mapping is an absolutely broken concept, and I don’t understand why after two years, this case’s behavior has changed to: https://github.com/up9cloud/config-rs/blob/master/tests/testsuite/file_jsonc.rs#L169.

I’m certain that when I submitted the pull request back then, all test cases passed.

This behavior should follow one clear approach, rather than just forcing lowercase:

- Flexible: env.`BAR`, env.`Bar`, and env.`bar` should all map to enum Settings { `Bar`(String) }.
- Strict: env.BAR should map only to BAR, env.bar should only map to bar, etc

polarathene:
> I’m certain that when I submitted the pull request back then, all test cases passed.

The feedback was back in 2023. You later pushed some commits in Feb 2024 and the PR was silent until recently. It's not a surprise that tests could fail given a new major zero ver release was made which permits breaking changes.

- Feb 2024 - [v14.0 introduced case insensitivity as a fix](https://github.com/rust-cli/config-rs/blob/master/CHANGELOG.md#0140---2024-02-01).
- Dec 2024 - [v15.0 reverted the case-insensitivity](https://github.com/rust-cli/config-rs/blob/master/CHANGELOG.md#0150---2024-12-17), there should be a related issue (or more) on the topic.

---

Regarding feedback with the `jsonc` feature concern for the `.json` extension, if it doesn't make sense to have other format features handle `.json` when `jsonc` is enabled, perhaps this could be detected as a conflict via `build.rs`? (_unless `Cargo.toml` can express this, [doesn't seem so](https://github.com/rust-lang/cargo/issues/2980)_)

I have seen crates like [`cudarc` handle it in `build.rs`](https://github.com/coreylowman/cudarc/blob/5163291f67d91ac9d3f758cd98d34f12394ba641/build.rs#L10-L15), so something like this?:

```rust
#[cfg(all(feature = "json", feature = "jsonc"))]
panic!("Both `json` and `jsonc` features are active, only one feature for parsing `.json` files is permitted");
```

---

Regarding `src/file/format/jsonc.rs`, what are your thoughts about [this proposed common method for serde supported formats](https://github.com/rust-cli/config-rs/pull/472#issuecomment-1770151034)? As the [`jsonc-parser` crate offers a `parse_to_serde_value()`](https://github.com/dprint/jsonc-parser#serde) method that returns a `serde_json::Value`, it should simplify the file in this PR to just:

```rust
use std::error::Error;

use crate::format;
use crate::map::Map;
use crate::value::Value;

pub fn parse(
    uri: Option<&String>,
    text: &str,
) -> Result<Map<String, Value>, Box<dyn Error + Send + Sync>> {
    // Parse a JSON/JSONC input from the provided text
    let value = format::from_parsed_value(uri, jsonc_parser::parse_to_serde_value(text)?);
    format::extract_root_table(uri, value)
}
```

`extract_root_table()` is already merged so formats could skip this portion of common logic:

https://github.com/rust-cli/config-rs/blob/07f13ff25f18e6a7bd40a2d8a2f1e036a473736a/src/format.rs#L27-L46

So that should already simplify your `parse()` method for `jsonc`, but as my PR introducing the `from_parsed_value()` method has been [recently questioned if it's worth merging with it's current approach](https://github.com/rust-cli/config-rs/pull/472#discussion_r2263128481), this PR would presently need to keep your `from_jsonc_value()` method.

The following should work for now at least:

```diff
-    let value = format::from_parsed_value(uri, jsonc_parser::parse_to_serde_value(text)?);
+    let value = from_jsonc_value(uri, jsonc_parser::parse_to_value(text, &Default::default())?);
    format::extract_root_table(uri, value)
```

Should #472 get merged, then `from_jsonc_value()` can also be dropped in favor of using `format::from_parsed_value()`.

---

**EDIT:** Oh just noticed that the return type of the result is the [value wrapped in an option](https://github.com/dprint/jsonc-parser/issues/52), which unlike `serde-json` will result in `""` returning `None` rather than an empty string. Going with the fallback you had instead:

```diff
-    let value = format::from_parsed_value(uri, jsonc_parser::parse_to_serde_value(text)?);
+    let parsed = jsonc_parser::parse_to_value(text, &Default::default())?;
+    let value = from_jsonc_value(uri, parsed.unwrap_or(ValueKind::Nil));
    format::extract_root_table(uri, value)
```

I'll apply that as a review suggestion 😅

src/file/format/jsonc.rs
```diff
@@ -0,0 +1,61 @@
...
+                ValueKind::Table(map) => Ok(map),
+                _ => unreachable!(),
+            },
+            JsonValue::Array(_) => Err(Unexpected::Seq),
+            JsonValue::Null => Err(Unexpected::Unit),
+        },
+        None => Err(Unexpected::Unit),
+    }
+    .map_err(|err| ConfigError::invalid_root(uri, err))
+    .map_err(|err| Box::new(err) as Box<dyn Error + Send + Sync>)
```
polarathene: This suggestion will require adding an import: `use crate::format;`

```suggestion
    let parsed = jsonc_parser::parse_to_value(text, &Default::default())?;
    let value = from_jsonc_value(uri, parsed.unwrap_or(ValueKind::Nil));
    format::extract_root_table(uri, value)
```

Cargo.toml
```diff
@@ -121,11 +121,12 @@ pre-release-replacements = [
...
 json = ["serde_json"]
 yaml = ["yaml-rust2"]
 ini = ["rust-ini"]
 json5 = ["json5_rs", "serde/derive"]
+jsonc = ["jsonc-parser"]
 convert-case = ["convert_case"]
 preserve_order = ["indexmap", "toml?/preserve_order", "serde_json?/preserve_order", "ron?/indexmap"]
```
polarathene: [Sep 2023 there was a comment about this feature not being supported by `jsonc-parser`](https://github.com/rust-cli/config-rs/pull/457#discussion_r1331132767), it [landed in Oct 2023](https://github.com/dprint/jsonc-parser/pull/39) as part of [`jsonc-parser@0.23.0`](https://github.com/dprint/jsonc-parser/releases/tag/0.23.0) so you'll want to [add the opt-in](
https://github.com/dprint/jsonc-parser/blob/cf46d1089aac3fe39b477e29f552fc208f0dd46a/Cargo.toml#L20) here?

```suggestion
preserve_order = ["indexmap", "toml?/preserve_order", "serde_json?/preserve_order", "jsonc-parser?/preserve_order", "ron?/indexmap"]
```

Cargo.toml
```diff
@@ -141,6 +142,7 @@ yaml-rust2 = { version = "0.10", optional = true }
 rust-ini = { version = "0.21", optional = true }
 ron = { version = "0.8", optional = true }
 json5_rs = { version = "0.4", optional = true, package = "json5" }
+jsonc-parser = { version = "0.23.0", optional = true }
```
polarathene:
```suggestion
jsonc-parser = { version = "0.26.0", optional = true }
```

Cargo.toml
polarathene:
```suggestion
default = ["toml", "json", "yaml", "ini", "ron", "json5", "convert-case", "async"]
```

epage:
Please rework your commits as meaningful, atomic commits to be merged.

up9cloud:
@epage, @polarathene all green now :D.

The .json vs .jsonc issue comes down to the project’s philosophy: should it aim to be flexible or strict?

If goes for flexible, the current behavior makes sense, because in practice, almost no one actually uses the .jsonc extension. People generally just want .json files that can contain comments.
If goes for strict, then only .jsonc should be allowed. Later, could consider adding a feature to configure which file extensions are parsed by which parser.

For example, from the earlier point about case insensitive handling and forcing lowercase mapping, these are closely tied to the overall design philosophy. Under a strict approach, case insensitive handling should be an optional feature rather than the default, and forcing lowercase mapping should be avoided - upper should map to upper, and lower should be lower.

epage:
From https://jsonc.org/

> The recommended file extension for JSONC documents is .jsonc.
>
> The extension .json should be avoided, but is supported if a mode line is present at the start of the file:

imo comments inside of json is no longer json.  We should be properly validating the json we parse.

Cargo.toml
```diff
@@ -126,8 +126,9 @@ json = ["serde_json"]
 yaml = ["yaml-rust2"]
 ini = ["rust-ini"]
 json5 = ["json5_rs", "serde/derive"]
+jsonc = ["jsonc-parser"]
```
epage:
```suggestion
jsonc = ["dep:jsonc-parser"]
```
#676 is tracking fixing the rest.

tests/testsuite/file_ini.rs
```diff
@@ -252,3 +252,16 @@ fn ini() {
...
+    let res = Config::builder()
+        .add_source(File::from_str("", FileFormat::Ini))
+        .build();
+    assert!(res.is_ok());
+    let c = res.unwrap();
+    assert_data_eq!(
+        format!("{:?}", c),
+        str!("Config { defaults: {}, overrides: {}, sources: [], cache: Value { origin: None, kind: Table({}) } }")
+    );
+}
```
epage: Why are these `test_nothing`s added?  if they are indirectly related (you wanted a `test_nothing` for jsonc), then please add these in their own commit before this one

tests/testsuite/file_ini.rs
```diff
@@ -252,3 +252,16 @@ fn ini() {
...
+
+#[test]
+fn test_nothing() {
+    let res = Config::builder()
+        .add_source(File::from_str("", FileFormat::Ini))
+        .build();
+    assert!(res.is_ok());
+    let c = res.unwrap();
+    assert_data_eq!(
+        format!("{:?}", c),
```
epage: `c.to_debug()`

tests/testsuite/file_jsonc.rs
```diff
@@ -0,0 +1,205 @@
...
+
+    assert!(res.is_err());
+    let err = res.unwrap_err().to_string();
+    let expected_prefix =
+        "Expected colon after the string or word in object property on line 4 column 1 in ";
+    assert!(
+        err.starts_with(expected_prefix),
+        "Error message does not start with expected prefix. Got: {}",
+        err
+    );
```
epage, line 108: Can we switch this to snapshot testing (`assert_data_eq!`) like our other formats?

tests/testsuite/file_ini.rs
polarathene: They might have added due to [an observation I also noticed](https://github.com/dprint/jsonc-parser/issues/52) with how `jsonc-parser` and `serde-json` differ in parsing an empty string (_`Ok(None)` vs `Error(E)`_).

So this might be a sanity check that on the `config-rs` side we're aware of that parsing difference? Especially for the `.json` extension [since this PR adopts that extension](https://github.com/rust-cli/config-rs/pull/457/files#r1331145558) in addition to the `serde-json` format parser.

---

I had previously suggested considering a [`build.rs` feature conflict check](https://github.com/rust-cli/config-rs/pull/457#issuecomment-3173241589), so that `json` vs `jsonc` are mutually exclusive? (_**EDIT:** actually may not be necessary since a [format hint can be provided](https://github.com/rust-cli/config-rs/blob/5f308736784460440f438b52296edc2b8e650102/src/file/source/file.rs#L25) to control selection, or [similar methods that take an explicit format](https://github.com/rust-cli/config-rs/blob/5f308736784460440f438b52296edc2b8e650102/src/file/mod.rs#L41-L77) arg_)

tests/testsuite/file_jsonc.rs
up9cloud: It was originally written using snapshot testing (assert_data_eq!), but it failed on Windows MSRV due to \ vs / path differences. Adding OS-specific handling would fix it, but it would make the code noticeably messier.

tests/testsuite/file_ini.rs
up9cloud: @polarathene is correct. since parsing results in None and config-rs does not have a corresponding type for that, these tests were added because .jsonc needs this case. The goal was to decide whether None should be treated as an error or ignored, and to compare behavior with other parsers.
That said, your point makes sense. I’ll remove the tests for the other parsers and keep only the .jsonc one.

up9cloud:
> From https://jsonc.org/
> 
> > The recommended file extension for JSONC documents is .jsonc.
> > The extension .json should be avoided, but is supported if a mode line is present at the start of the file:
> 
> imo comments inside of json is no longer json. We should be properly validating the json we parse.

This ultimately comes down to a design philosophy question. While the spec says otherwise, in practice, especially for configuration files, almost nobody uses the .jsonc as extension; all with .json.

That said, this discussion could go on forever. For now, I’ll change it so that jsonc only parses .jsonc. If there’s a decision later to move toward a more user-friendly approach, we can revisit it.

tests/testsuite/file_ini.rs
epage: For existing formats can you add the tests in a separate commit before jsonc support as mentioned earlier in this thread?

up9cloud: I’d like to, but preferably after this current pull request is resolved. Since it’s been almost two years of inactivity on this thread, focusing on closing this one first makes the process clearer and prevents scope creep. Afterward, I can gradually add more test cases, including for the existing formats.

up9cloud:
I accidentally clicked “Review changes” earlier but forgot to press the “Submit review” button, so my comment was left in pending status. It should be visible now.

PR #460, opened 2023-09-25
Use weak features for preserve_order

sunshowers:
Currently, `preserve_order` causes the toml, serde_json and ron dependencies to be enabled. This can be avoided by using cargo's support for [weak features].

[weak features]:
https://rust-lang.github.io/rfcs/3143-cargo-weak-namespaced-features.html

Weak features were stabilized in Rust 1.60 so there are no MSRV concerns.

I think this probably is a breaking change since it's possible some downstream users weren't enabling support explicitly.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,7 +21,7 @@ yaml = ["yaml-rust"]
 ini = ["rust-ini"]
 json5 = ["json5_rs", "serde/derive"]
 convert-case = ["convert_case"]
-preserve_order = ["indexmap", "toml/preserve_order", "serde_json/preserve_order", "ron/indexmap"]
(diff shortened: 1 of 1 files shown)


matthiasbeyer (approved):
I like it!

I agree that this is a breaking change, but I don't think that's too much of an issue as we're in 0.y.z anyways, so that'll only cause a y+1 release!

Thanks for your contribution!

matthiasbeyer:
Unfortunately, CI fails because your commit needs signoff. Please:

```
git commit --amend --signoff && git push <your remote> weak --force
```

to fix this.

sunshowers:
@matthiasbeyer done, thanks!

Issue #461, opened 2023-09-27
Cannot deserialize with Struct without fields 


zhongzc:
```rust
#[derive(Debug, Serialize, Deserialize)]
struct A {
    b: Option<B>,
}
#[derive(Debug, Serialize, Deserialize)]
struct B {}

let a = A { b: Some(B {}) };
let de: A = Config::try_from(&a).unwrap().try_deserialize().unwrap();
assert!(de.b.is_some()); // Failed
```

matthiasbeyer:
How would you specify that in a configuration file?

matthiasbeyer:
And again I ask: How would you specify a instance of `A { b: Some(B) }` in a configuration file?

zhongzc:
Just like the previous example, for toml file, I specify it like this:
```toml
[b]
```

matthiasbeyer:
Ah, I missed that bit. Hnm, indeed that's an issue. Thanks for reporting!

matthiasbeyer:
I opened #463 with a testcase for this issue. If you have an idea solving that problem, you're of course very much welcome!

zhongzc:
The scenarios that cause `B` to be lost are related to iterators. An example that better reflects the situation I found:

```rust
#[derive(Debug, Serialize, Deserialize, PartialEq)]
struct A {
    b_option: Option<B>,
    b_vec: Vec<B>,
    b_set: HashSet<B>,
    b_map: HashMap<String, B>,
}
#[derive(Debug, Serialize, Deserialize, PartialEq, Eq, Hash)]
struct B {}

let a = A {
    b_option: Some(B {}),
    b_vec: vec![B {}],
    b_set: HashSet::from([B {}]),
    b_map: HashMap::from([("".to_owned(), B {})]),
};

// Serialize that `a` with `ConfigSerializer` will output an empty config
let config = Config::try_from(&a).unwrap();
```

polarathene:
**TL;DR:** Looks like this can be resolved easily enough with the below bolded fixes.

---

Note that there is two ways to define a unit-like struct (_[technically more](https://stackoverflow.com/questions/50162597/what-are-the-differences-between-the-multiple-ways-to-create-zero-sized-structs/50170328#50170328)?_):

```rs
// Officially documented way:
// https://doc.rust-lang.org/book/ch05-01-defining-structs.html#unit-like-structs-without-any-fields
// https://learning-rust.github.io/docs/structs/
struct B;

// RFC:
[...]

PR #463, opened 2023-09-28
Add test for empty inner object

matthiasbeyer:
So far this only adds a test for an empty inner object as reported in #461.

If someone wants to help solving that issue, you're very much welcome.


polarathene:
Possible fix identified: https://github.com/mehcode/config-rs/issues/461#issuecomment-1771963758

PR #470, opened 2023-10-06
refactor: `ini` format parser

polarathene:
Intended to be easier to grok and match the same flow as other format parsers.

---

I worked on this after https://github.com/mehcode/config-rs/pull/469 , but my main gripes were with the `sec` not being obvious that it was referencing a section, and the repeated inner loop for building property maps.

I added the same enum check with the final `Table` map wrapped in `Value` like other parsers, although for this that may be redundant and could be dropped?

Not too fussed if you want to reject this. I originally envisioned it looking nicer than it turned out 😅


src/file/format/ini.rs
```diff
@@ -3,35 +3,45 @@ use std::error::Error;
...
-                }
-            }
-        }
+    let value = from_ini(uri, Ini::load_from_str(text)?);
+
+    match value.kind {
+        ValueKind::Table(map) => Ok(map),
+
+        _ => Ok(Map::new()),
+    }
```
matthiasbeyer, line 18: How about something like

```rust
if let ValueKind::Table(map) = value.kind {
    Ok(map)
} else {
    Ok(Map::new)
}
```

here?

src/file/format/ini.rs
```diff
@@ -3,35 +3,45 @@ use std::error::Error;
...
+    data: Ini,
+) -> Value {
+    let mut map = Map::new();
+
+    let mut sections: Map<Option<&str>, Table> = data.into_iter().map(|(section, props)| {(
+        section,
+        props.iter().map(|(k, v)| {(
+            k.to_owned(),
+            Value::new(uri, ValueKind::String(v.to_owned())),
+        )}).collect()
```
matthiasbeyer, line 32: This is really unreadable.

How about
```suggestion
        props.iter().map(|(k, v)| {
            let key = k.to_owned();
            let value = Value::new(uri, ValueKind::String(v.to_owned());
            (key, value)
        }).collect()
```

instead of the rather hard to parse tuple inside the block?

src/file/format/ini.rs
```diff
@@ -3,35 +3,45 @@ use std::error::Error;
...
+    if let Some(sectionless) = sections.remove(&None) {
+        map.extend(sectionless);
     }
-    Ok(map)
+
+    // Wrap each section Table into Value for merging into `map`:
+    map.extend(sections.into_iter().map(|(k,v)| {(
+        k.unwrap_or_default().to_owned(),
+        Value::new(uri, ValueKind::Table(v)),
+    )}));
```
matthiasbeyer, line 44: Please expand the tuple in the same way as suggested above.

matthiasbeyer (changes requested):

src/file/format/ini.rs
polarathene: I used what you have for every other format atm. A separate PR I referenced provides a common method for this logic based on a TODO you have in each of the files.

Thus if you approve of that approach, it'd be better to just rebase here and switch to the method?

matthiasbeyer: Agreed

src/file/format/ini.rs
polarathene: Yeah that's much better, thanks for the suggestion! :)

src/file/format/ini.rs
```diff
@@ -3,35 +3,45 @@ use std::error::Error;
...
+        props.iter().map(|(k, v)| {(
+            k.to_owned(),
+            Value::new(uri, ValueKind::String(v.to_owned())),
+        )}).collect()
+    )}).collect();
+
+    // These (optional) properties should exist top-level alongside sections:
+    if let Some(sectionless) = sections.remove(&None) {
+        map.extend(sectionless);
     }
```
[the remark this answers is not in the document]
matthiasbeyer: > but they're optional, so I figured a conditional was necessary?

Well, `Option` is an iterator, so calling `into_iter()` on it either returns an iterator yielding one element (in the `Some` case) or no element (in the `None` case). So I _think_ this could work.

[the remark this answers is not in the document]
polarathene: Tried a few attempts with your suggestion but kept getting failures like:

image

However it seems to work well if I just unwrap the option? Tests are passing with the sectionless properties and I verified that everything works fine if there is no sectionless properties too.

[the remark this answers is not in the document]
matthiasbeyer: Hm okay. Unwrapping is not an option if we can implement it without unwrapping.

src/file/format/ini.rs
```diff
@@ -22,26 +22,26 @@ fn from_ini(
...
+            (key, value)
+        }).collect()
     )}).collect();
 
-    // These (optional) properties should exist top-level alongside sections:
-    if let Some(sectionless) = sections.remove(&None) {
-        map.extend(sectionless);
-    }
+    // Hoist (optional) sectionless properties to the top-level, alongside sections:
+    map.extend(sections.remove(&None).unwrap_or_default());
```
matthiasbeyer, line 37: Ah, a safe unwrap. Okay if that works, I'd be fine.

src/file/format/ini.rs
```diff
@@ -3,35 +3,45 @@ use std::error::Error;
...
+    let mut map = Map::<String, Value>::new();
+
+    let mut sections: Map<Option<&str>, Table> = data.into_iter().map(|(section, props)| {(
+        section,
+        props.iter().map(|(k, v)| {
+            let key = k.to_owned();
+            let value = Value::new(uri, ValueKind::String(v.to_owned()));
+            (key, value)
+        }).collect()
+    )}).collect();
```
matthiasbeyer, line 34: Here we have an iterator inside an iterator. I bet this could be rewritten to something like:

```suggestion
    let mut sections: Map<Option<&str>, Table> = data.into_iter().flat_map(|(section, props)| {(
        section,
        props.iter().map(|(k, v)| {
            let key = k.to_owned();
            let value = Value::new(uri, ValueKind::String(v.to_owned()));
            (key, value)
        })
    )}).collect();
```

Also I don't like the outer tuple (didn't notice before), so that could be expanded as well.

Sorry for continuously bugging you about this piece :laughing:

matthiasbeyer (changes requested):

src/file/format/ini.rs
polarathene: > Here we have an iterator inside an iterator. I bet this could be rewritten to something like:

I don't think [FlatMap](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.flat_map) is going to work. We are collecting like this:

```
sectionless_property=value

[section_a]
property_in_section=value
```

Each section has it's own collection of properties, while the config is a collection of sections (_with the sectionless one hoisted up_). The nesting is necessary? At an earlier iteration, I had extracted the inner collection for properties to a separate function call, I could switch to that if you find it more readable.

> Also I don't like the outer tuple (didn't notice before), so that could be expanded as well.

No worries, I'll take care of that 👍 

> Sorry for continuously bugging you about this piece 😆

It's alright, I'm the same when I review contributions too 😛

polarathene (commented):
That's me for the night, I'll address any remaining feedback tomorrow 😴

src/file/format/ini.rs
matthiasbeyer: > I don't think [FlatMap](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.flat_map) is going to work. We are collecting like this:

Ah, now I see it that there's a tuple as well! That's an oversight at my side, sorry. Sure, then `flat_map` does not work here.

PR #472, opened 2023-10-07
refactor: Use a common `from_value` method

polarathene:
Inspired from the `json5` feature (_Original PR July 2020, revised PR May 2021_) with the refactoring in the revision by @matthiasbeyer 

It looked useful for other formats that use `serde` to simplify their logic (_so long as test coverage is informative of no issues_ 🙏 )

---

DRY: The `json`, `json5` and `toml` parsers all leverage `serde` and can share a common enum to deserialize data into, instead of individual methods performing roughly the same transformations.

- While `ron` improves their support for serde untagged enums with v0.9, ~~it is still not compatible with this approach (_Their README details why_).~~ (_**UPDATE:** At least for the current `config-rs` test coverage, this appears to be passing now_)
- The `yaml` support doesn't leverage `serde` thus is not compatible. (_**UPDATE:** Since there is talk about the unmaintained status, it could be switched for [`serde-yaml`](https://github.com/dtolnay/serde-yaml), I've verified it can pass the tests with an extra deserializer method_)

`from_parsed_value()` is based on the approached used by `format/json5.rs`:
- It has been adjusted to reflect the `ValueKind` enum, which could not directly be used due to the `Table` and `Array` types using `Value` as their value storage type instead of self-referencing the enum.
- Very similar to a `impl From`, but supports the complimentary `uri` parameter for each `Value` derived.

---

Resolves: https://github.com/mehcode/config-rs/pull/394

### DIFF
--- a/src/format.rs
+++ b/src/format.rs
@@ -44,3 +44,53 @@ pub fn extract_root_table(
     .map_err(|err| ConfigError::invalid_root(uri, err))
     .map_err(|err| Box::new(err) as Box<dyn Error + Send + Sync>)
 }
+
+// Equivalent to ValueKind, except Table + Array store the same enum
+// Useful for serde to serialize values into, then convert to Value
+#[derive(serde::Deserialize, Debug)]
+#[serde(untagged)]
+pub enum ParsedValue {
+    Nil,
+    Boolean(bool),
+    I64(i64),
+    I128(i128),
+    U64(u64),
+    U128(u128),
+    Float(f64),
+    String(String),
+    Table(Map<String, Self>),
+    Array(Vec<Self>),
+}
+
+// Value wrap ValueKind values, with optional uri (origin)
+pub fn from_parsed_value(uri: Option<&String>, value: ParsedValue) -> Value {
+    let vk = match value {
+        ParsedValue::Nil => ValueKind::Nil,
+        ParsedValue::String(v) => ValueKind::String(v),
+        ParsedValue::I64(v) => ValueKind::I64(v),
+        ParsedValue::I128(v) => ValueKind::I128(v),
+        ParsedValue::U64(v) => ValueKind::U64(v),
+        ParsedValue::U128(v) => ValueKind::U128(v),
+        ParsedValue::Float(v) => ValueKind::Float(v),
+        ParsedValue::Boolean(v) => ValueKind::Boolean(v),
+        ParsedValue::Table(table) => {
+            let m = table
+                .into_iter()
+                .map(|(k, v)| (k, from_parsed_value(uri, v)))
+                .collect();
+
+            ValueKind::Table(m)
+        }
+
+        ParsedValue::Array(array) => {
+            let l = array
+                .into_iter()
+                .map(|v| from_parsed_value(uri, v))
+                .collect();
+
+            ValueKind::Array(l)
+        }
+    };
+
+    Value::new(uri, vk)
+}
--- a/src/format.rs
+++ b/src/format.rs
@@ -3,6 +3,7 @@ use std::error::Error;
 use crate::error::{ConfigError, Unexpected};
 use crate::map::Map;
 use crate::value::{Value, ValueKind};
+use serde::Deserialize;
 
 /// Describes a format of configuration source data
 ///
@@ -63,34 +64,6 @@ pub enum ParsedValue {
     Array(Vec<Self>),
 }
 
-// Deserialization support for TOML `Datetime` value type into `String`
-#[derive(serde::Deserialize, Debug)]
-#[serde(untagged)]
-enum ParsedString {
-    String(String),
-    #[cfg(feature = "toml")]
-    DateTime(toml::value::Datetime),
-}
-
-fn deserialize_parsed_string<'de, D>(deserializer: D) -> Result<String, D::Error>
-where
-    D: serde::de::Deserializer<'de>,
-{
-    let s: ParsedString = serde::Deserialize::deserialize(deserializer)?;
-    Ok(s.to_string())
-}
-
-impl std::fmt::Display for ParsedString {
-    fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
-        let s = match self {
-            ParsedString::String(s) => s.to_string(),
-            #[cfg(feature = "toml")]
-            ParsedString::DateTime(dt) => dt.to_string()
-        };
-        write!(f, "{}", s)
-    }
-}
-
 // Value wrap ValueKind values, with optional uri (origin)
 pub fn from_parsed_value(uri: Option<&String>, value: ParsedValue) -> Value {
     let vk = match value {
@@ -122,4 +95,26 @@ pub fn from_parsed_value(uri: Option<&String>, value: ParsedValue) -> Value {
     };
 
     Value::new(uri, vk)
+}
+
+// Deserialization support for TOML `Datetime` value type into `String`
+fn deserialize_parsed_string<'de, D>(deserializer: D) -> Result<String, D::Error>
+where
+    D: serde::de::Deserializer<'de>,
+{
+    #[derive(serde::Deserialize)]
+    #[serde(untagged)]
+    enum ParsedString {
+        // Anything that can deserialize into a string successfully:
+        String(String),
+        // Config specific support for types that need string conversion:
+        #[cfg(feature = "toml")]
+        TomlDateTime(toml::value::Datetime),
+    }
+
+    Ok(match ParsedString::deserialize(deserializer)? {
+        ParsedString::String(v) => v,
+        #[cfg(feature = "toml")]
+        ParsedString::TomlDateTime(v) => v.to_string(),
+    })
 }
--- a/src/format.rs
+++ b/src/format.rs
@@ -121,23 +121,16 @@ where
         // Anything that can deserialize into a string successfully:
         String(String),
         // Config specific support for types that need string conversion:
+        Char(char),
         #[cfg(feature = "toml")]
         TomlDateTime(toml::value::Datetime),
-        #[cfg(feature = "ron")]
-        RonChar(ron::Value),
     }
 
     match ParsedString::deserialize(deserializer)? {
         ParsedString::String(v) => Ok(v),
+        ParsedString::Char(v) => Ok(v.to_string()),
         #[cfg(feature = "toml")]
         ParsedString::TomlDateTime(v) => Ok(v.to_string()),
-        #[cfg(feature = "ron")]
-        ParsedString::RonChar(variant) => match variant {
-            ron::Value::Char(v) => Ok(v.to_string()),
-            _ => Err(serde::de::Error::custom(
-                "should not be serialized to string",
-            )),
-        },
     }
 }
 
--- a/src/format.rs
+++ b/src/format.rs
@@ -4,6 +4,7 @@ use crate::error::{ConfigError, Unexpected};
 use crate::map::Map;
 use crate::value::{Value, ValueKind};
 use serde::Deserialize;
+use serde_with::rust::deserialize_ignore_any;
 
 /// Describes a format of configuration source data
 ///
@@ -47,11 +48,12 @@ pub fn extract_root_table(
 }
 
 // Equivalent to ValueKind, except Table + Array store the same enum
-// Useful for serde to serialize values into, then convert to Value
+// Useful for serde to serialize values into, then convert to Value.
+// NOTE: Order of variants is important. Serde will use whichever
+// the input successfully deserializes into first.
 #[derive(serde::Deserialize, Debug)]
 #[serde(untagged)]
 pub enum ParsedValue {
-    Nil,
     Boolean(bool),
     I64(i64),
     I128(i128),
@@ -62,6 +64,9 @@ pub enum ParsedValue {
     String(String),
     Table(Map<String, Self>),
     Array(Vec<Self>),
+    // If nothing else above matched, use Nil:
+    #[serde(deserialize_with = "deserialize_ignore_any")]
+    Nil,
 }
 
 // Value wrap ValueKind values, with optional uri (origin)
--- a/src/format.rs
+++ b/src/format.rs
@@ -62,6 +62,7 @@ pub enum ParsedValue {
     Float(f64),
     #[serde(deserialize_with = "deserialize_parsed_string")]
     String(String),
+    #[serde(deserialize_with = "deserialize_parsed_map")]
     Table(Map<String, Self>),
     Array(Vec<Self>),
     Option(Option<Box<Self>>),
@@ -137,5 +138,44 @@ where
                 "should not be serialized to string",
             )),
         },
+    }
+}
+
+fn deserialize_parsed_map<'de, D>(deserializer: D) -> Result<Map<String, ParsedValue>, D::Error>
+where
+    D: serde::de::Deserializer<'de>,
+{
+    #[derive(serde::Deserialize)]
+    #[serde(untagged)]
+    enum ParsedMap {
+        // Anything that can deserialize into a Map successfully:
+        Table(Map<String, ParsedValue>),
+        // Config specific support for types that need string conversion:
+        #[cfg(feature = "yaml")]
+        YamlMap(serde_yaml::Mapping),
+    }
+
+    match ParsedMap::deserialize(deserializer)? {
+        ParsedMap::Table(v) => Ok(v),
+        #[cfg(feature = "yaml")]
+        ParsedMap::YamlMap(table) => {
+            table
+                .into_iter()
+                .map(|(key, value)| {
+                    let key = match key {
+                        serde_yaml::Value::Number(k) => Some(k.to_string()),
+                        serde_yaml::Value::String(k) => Some(k),
+                        _ => None,
+                    };
+                    let value = serde_yaml::from_value::<ParsedValue>(value).ok();
+
+                    // Option to Result:
+                    match (key, value) {
+                        (Some(k), Some(v)) => Ok((k, v)),
+                        _ => Err(serde::de::Error::custom("should not be serialized to Map")),
+                    }
+                })
+                .collect()
+        }
     }
 }
--- a/src/format.rs
+++ b/src/format.rs
@@ -64,6 +64,7 @@ pub enum ParsedValue {
     String(String),
     Table(Map<String, Self>),
     Array(Vec<Self>),
+    Option(Option<Box<Self>>),
     // If nothing else above matched, use Nil:
     #[serde(deserialize_with = "deserialize_ignore_any")]
     Nil,
@@ -80,6 +81,7 @@ pub fn from_parsed_value(uri: Option<&String>, value: ParsedValue) -> Value {
         ParsedValue::U128(v) => ValueKind::U128(v),
         ParsedValue::Float(v) => ValueKind::Float(v),
         ParsedValue::Boolean(v) => ValueKind::Boolean(v),
+
         ParsedValue::Table(table) => {
             let m = table
                 .into_iter()
@@ -97,12 +99,17 @@ pub fn from_parsed_value(uri: Option<&String>, value: ParsedValue) -> Value {
 
             ValueKind::Array(l)
         }
+
+        // Boxed value must be dereferenced:
+        ParsedValue::Option(v) => match v {
+            Some(boxed) => from_parsed_value(uri, *boxed).kind,
+            None => ValueKind::Nil,
+        },
     };
 
     Value::new(uri, vk)
 }
 
-// Deserialization support for TOML `Datetime` value type into `String`
 fn deserialize_parsed_string<'de, D>(deserializer: D) -> Result<String, D::Error>
 where
     D: serde::de::Deserializer<'de>,
@@ -115,11 +122,20 @@ where
         // Config specific support for types that need string conversion:
         #[cfg(feature = "toml")]
         TomlDateTime(toml::value::Datetime),
+        #[cfg(feature = "ron")]
+        RonChar(ron::Value),
     }
 
-    Ok(match ParsedString::deserialize(deserializer)? {
-        ParsedString::String(v) => v,
+    match ParsedString::deserialize(deserializer)? {
+        ParsedString::String(v) => Ok(v),
         #[cfg(feature = "toml")]
-        ParsedString::TomlDateTime(v) => v.to_string(),
-    })
+        ParsedString::TomlDateTime(v) => Ok(v.to_string()),
+        #[cfg(feature = "ron")]
+        ParsedString::RonChar(variant) => match variant {
+            ron::Value::Char(v) => Ok(v.to_string()),
+            _ => Err(serde::de::Error::custom(
+                "should not be serialized to string",
+            )),
+        },
+    }
 }
--- a/src/format.rs
+++ b/src/format.rs
@@ -57,9 +57,38 @@ pub enum ParsedValue {
     U64(u64),
     U128(u128),
     Float(f64),
+    #[serde(deserialize_with = "deserialize_parsed_string")]
     String(String),
     Table(Map<String, Self>),
     Array(Vec<Self>),
+}
+
+// Deserialization support for TOML `Datetime` value type into `String`
+#[derive(serde::Deserialize, Debug)]
+#[serde(untagged)]
+enum ParsedString {
+    String(String),
+    #[cfg(feature = "toml")]
+    DateTime(toml::value::Datetime),
+}
+
+fn deserialize_parsed_string<'de, D>(deserializer: D) -> Result<String, D::Error>
+where
+    D: serde::de::Deserializer<'de>,
+{
+    let s: ParsedString = serde::Deserialize::deserialize(deserializer)?;
+    Ok(s.to_string())
+}
+
+impl std::fmt::Display for ParsedString {
+    fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
+        let s = match self {
+            ParsedString::String(s) => s.to_string(),
+            #[cfg(feature = "toml")]
+            ParsedString::DateTime(dt) => dt.to_string()
+        };
+        write!(f, "{}", s)
+    }
 }
 
 // Value wrap ValueKind values, with optional uri (origin)
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,7 +17,7 @@ maintenance = { status = "actively-developed" }
 [features]
 default = ["toml", "json", "yaml", "ini", "ron", "json5", "convert-case", "async"]
 json = ["serde_json"]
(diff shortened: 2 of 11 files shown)


polarathene:
I was able to resolve the datetime error by including the toml type for it in the `ParsedValue` enum, but since that'd interfere with the generic method and the feature conditional compilation, I guess I should revert the `toml.rs` change.

It's Datetime is not compatible with chrono DateTime type unfortunately, at least serde doesn't recognize it for that.

polarathene:
This PR has been updated to also include support for formats Ron and Yaml.
- They're split into their own commits those can be easily dropped (_and split into follow-up PRs if preferred, especially for the Yaml crate switch_).
- Alternative PRs have been provided that could optionally be merged before this PR, or if any portion of this PR is rejected.

Ron:
  - Specific commit: https://github.com/mehcode/config-rs/pull/472/commits/e597f6aaef1205d508480735bf887bb8301408d9
  - Alternative PR (_without common `from_value`_): https://github.com/mehcode/config-rs/pull/476

Yaml:
  - Specific commit: https://github.com/mehcode/config-rs/pull/472/commits/d4f2f354fc0a78eaf7b77e5ad528d7c11b475961
  - Alternative PR (_without common `from_value`_): https://github.com/mehcode/config-rs/pull/474

---

During the integration of Ron and Yaml into this PR (_like with earlier TOML test for datetime_), both failed running their tests due to the more implicit `deserialization_any` used by Serde to support deserializing for the untagged enum.

Resolving requires either:
  - Adding a new type to deserialize into for `from_value()` to then handle
  - Using the serde `with` attribute to provide a custom deserializer that better guides handling of each formats' types into the expected type for `config-rs` (_eg: Ron `Char` or Toml `Datetime` conversion to `String`_).

It's not too different from the prior approach, these types were in the formats own `from_value()` method, whereas this approach better identifies when this is actually necessary. They still leverage feature toggles for conditional compilation.

---

For TOML with the `Datetime` type that was easier to resolve thanks to the specific test failure. Ron and YAML failures were more cryptic usually, such as:

```
running 7 tests
test test_error_parse ... ok
test test_override_lowercase_value_for_enums ... ok
test test_override_uppercase_value_for_enums ... ok
test test_override_lowercase_value_for_struct ... ok
test test_override_uppercase_value_for_struct ... ok
test test_file ... ok

thread 'test_yaml_parsing_key' has overflowed its stack
fatal runtime error: stack overflow
error: test failed, to rerun pass `--test file_yaml`
```

Which IIRC I had read was due to how `deserialize_any` works with the untagged enum, that is a situation of it recursively trying to resolve the enum incorrectly due to reaching a variant that references `Self` and performing `deserialize_any` on that and so forth.
- That's the main maintenance gotcha, but I doubt it would be encountered much going forward, and this approach simplifies the maintenance as a whole quite well.
- `deserialize_any` (_due to untagged enum approach_) won't be as performant as the previous `match` statements for each individual format, but unlikely a concern for the purpose of `config-rs`.

---

> **Note**
> Ron `0.9` will introduce a Bytes value variant (`Vec<u8>`) for byte strings.
> - I don't believe the current approach would match it? You'd get the fallback `Nil`.
> - Probably acceptable given it's unclear how `config-rs` would want to support that.

polarathene (commented):
Some inline context if it assists with review.

If it's easier to follow/digest, I've staged out changes into [scoped commits](https://github.com/mehcode/config-rs/pull/472/commits). Each commit has an associated commit message describing the changes 👍

polarathene:
@matthiasbeyer this is ready for review, I know you're quite busy so here's a TLDR:
- 🎉 All currently supported formats (_except INI_) leverage serde to reduce maintenance burden:
  - Supersedes: https://github.com/mehcode/config-rs/pull/474 + https://github.com/mehcode/config-rs/pull/476
  - Better prevents issues like: https://github.com/mehcode/config-rs/pull/394
- 🚀 Greatly simplifies adding new formats [that provide serde support](https://docs.rs/serde/latest/serde/#data-formats):
  - https://github.com/mehcode/config-rs/pull/466 (_adjusted to leverage this PR_)
  - https://github.com/mehcode/config-rs/pull/467 (_prior more verbose approach_)
  - I've implemented jsonc, hjson and dotenv locally with minimal effort.
- 😕 [This custom format example comment](https://github.com/mehcode/config-rs/blob/55c464e143d7301f3c107a405827fe9dc085f81a/examples/custom_format/main.rs#L27) won't make much sense after this PR.
  - It could [add a link to a past release](https://github.com/mehcode/config-rs/tree/0.13.3/src/file/format)? (_although most formats are roughly the same as the unified `from_value()` here, this PR approach adds another level of indirection with `ParsedValue`_).
  - I've successfully verified serde format crates Gura, Dhall, Hjson work well as custom formats when paired with this PR `from_value()` method (_no changes needed_).

polarathene:
For custom format support, the `format` method added here, one I previously contributed (`extract_root_table()`) and sometimes `ParsedValue` are required to leverage it (_to call `ParsedValue::deserialize(input_value)`_). This would require `lib.rs` to make the module `pub`, or some refactoring.

As it's all focused on using Serde with untagged enum, perhaps it shouldn't live in `format.rs` if it isn't useful to other formats (_not that I expect those to be contributed much_).

---

I also need to verify an assumption that other number types like `u8` would correctly be cast to `u64` supported by `config-rs`. Prior to this the mapping was more explicit, and I may have a misunderstanding with the untagged enum approach using `deserialize_any()` this way 😅 

I will also share the other formats I integrated locally, if desired they can be contributed via follow-up PR.

polarathene:
> After that, I think you can merge this!

I still need to address my concerns from my prior comment here. I will return to `config-rs` later today or tomorrow hopefully, presently tied up elsewhere for a bit.

After that's sorted locally, I'll rebase as requested and it'd be great to see this get merged :)

Regarding custom format support, how would you approach verifying that? Mostly so that you can catch that `format` module should be made public, incase someone were to revert that accidentally?

polarathene:
Note to self, since this PR was raised the proposed yaml alternative (serde compatible) has been archived and a different yaml crate has been introduced instead, so this PR will need to drop that support (_unfortunate, since the serde approach was preferrable I think_).

src/format.rs
```diff
@@ -44,3 +46,129 @@ pub fn extract_root_table(
     .map_err(|err| ConfigError::invalid_root(uri, err))
     .map_err(|err| Box::new(err) as Box<dyn Error + Send + Sync>)
 }
+
+// Equivalent to ValueKind, except Table + Array store the same enum
+// Useful for serde to serialize values into, then convert to Value.
+// NOTE: Order of variants is important. Serde will use whichever
+// the input successfully deserializes into first.
+#[derive(serde::Deserialize, Debug)]
+#[serde(untagged)]
```
epage, line 55: `serde(untagged)` is a performance footgun in situations like this (easily noticeable in benchmarks when I made the same mistake in `toml::Value`).

We'd likely want to use `serde-untagged`.

Yes, json5 did this already but I care more if it spreads to all formats

src/format.rs
```diff
@@ -44,3 +46,129 @@ pub fn extract_root_table(
     .map_err(|err| ConfigError::invalid_root(uri, err))
     .map_err(|err| Box::new(err) as Box<dyn Error + Send + Sync>)
 }
+
+// Equivalent to ValueKind, except Table + Array store the same enum
+// Useful for serde to serialize values into, then convert to Value.
+// NOTE: Order of variants is important. Serde will use whichever
+// the input successfully deserializes into first.
```
epage, line 53: As this just replaces our existing type we are deserializing to, I'll give this a pass.  Overall, I'd rather we deserialize directly into our needed type.

That would either require
- Putting a dummy value in for the origin
- Building up my `serde-basepath` idea I've been toying with, see https://github.com/clap-rs/clap/discussions/2763#discussioncomment-13915276

src/format.rs
polarathene: > `serde(untagged)` is a performance footgun in situations like this (easily noticeable in benchmarks when I made the same mistake in `toml::Value`).

Ah, I think when I contributed this I favored it for being more convenient maintenance wise and thinking of it from an application config viewpoint where you'd usually not have so much config to parse that it'd be a perf concern? 😅

> We'd likely want to use `serde-untagged`.

Glancing over at the [crate docs](https://docs.rs/serde-untagged/latest/serde_untagged/), I assume this means losing the convenience of `serde(untagged)` and requires a bit more explicit mapping to each variant?

I don't have a problem with that, but it was nice to have for simple mappings of types to their variant.

---

> Yes, json5 did this already but I care more if it spreads to all formats

This wasn't clear to me at first where you were referencing as I thought you meant upstream `json5` crate (_that only [uses `serde(untagged)` in their tests](https://github.com/search?q=repo%3Acallum-oakley%2Fjson5-rs%20untagged&type=code)_) 😅 but I see you meant the `config-rs` [format implementation for `json5`](https://github.com/rust-cli/config-rs/blob/07f13ff25f18e6a7bd40a2d8a2f1e036a473736a/src/file/format/json5.rs#L7-L17) uses `serde(untagged)` (_and is removed via this PR too_) 👍 (**EDIT:** I see it's also the first thing my PR description mentioned, I forgot as it's been a while 😓)

So as a bonus perhaps, this would address your concern about the json5 format being deferred to this common serde handler for supported formats?

src/format.rs
polarathene: > As this just replaces our existing type we are deserializing to, I'll give this a pass. Overall, I'd rather we deserialize directly into our needed type.

I believe that I wanted to deserialize directly too, but I think the comment here highlighted the reasoning that did not happen was due to [`config::Value` being a struct with a `value: ValueKind` field](https://github.com/rust-cli/config-rs/blob/07f13ff25f18e6a7bd40a2d8a2f1e036a473736a/src/value.rs#L184-L207) while [`config::ValueKind` enum has the `Table` + `Array` variants reference `config::Value` instead of `Self`](https://github.com/rust-cli/config-rs/blob/07f13ff25f18e6a7bd40a2d8a2f1e036a473736a/src/value.rs#L15-L30) to self-reference the enum, which complicated adapting to that?

If you look at what this PR does, it effectively replaces the format specific methods (_like `from_yaml_value()` / `from_toml_value()`_) with a single `from_parsed_value()` method instead.

Most of those format `from_*_value()` methods (_that are serde reliant_) took in a deserialized format value type as a parameter, delegating away error handling involved (_except partially for Ron and the non-serde Yaml format_).

In both cases the before/after approach of these methods are only really performing the task mapping to `config::Value`. That logic doesn't belong in `value.rs` when it's format/type specific though? You can see at the end of `format.rs` in this PR how the additional serde deserializers are added for format specific support concerns, with types/crates not really relevant to bundle into `value.rs`?

---

Hopefully that clarifies why `ParsedValue` was introduced as an intermediary convenience for the deserializing + mapping step. With `from_parsed_value()` it effectively consolidates the associated formats it can support, making this logic in `config-rs` much easier to grok and maintain?

Last I recall the various `from_*_value()` implementations were (_or at least at some point when I was contributing_) not particularly consistent in style/quality despite having roughly the same logic. One could tidy that up, or just make it DRY (which this PR favored), such that any new serde compatible formats would be much simpler to add/support, and thus maintain without as much boilerplate prone to diverging.

Existing (and new) serde compatible formats would have this common call going forward:

```rust
let value = format::from_parsed_value(uri, serde_json::from_str(text)?);
```

---

If `format.rs` can be adapted to make needing `ParsedValue` redundant (_by adapting the same improvement to `config::Value` / `ValueKind` directly_), that'd be great too! This PR was just focused on that initial step to take a DRY approach without being too invasive of a change to the codebase 😅

That said some of the PR value was lost since as [the switch to `serde-yaml` didn't age well](https://github.com/rust-cli/config-rs/issues/675#issuecomment-3166247688). It would have been nice to shift the added error handling there to serde.

polarathene: > Putting a dummy value in for the origin

**NOTE:** I used `uri` in my responses since it's frequently the variable name used instead of `origin` and was what came to mind when writing out my thoughts 😅

---

`from_parsed_value()` (_like how other format methods before it did_) provides the `uri` value to use when mapping from a `ParsedValue` to a `config::Value` (_with it's associated `uri` + `ValueKind` variant_).

Adding a placeholder `uri` (effectively `None`?) would then need a step that updates the `uri` for each `Value` (_along with it's nested `Value` items depending on type_), and I'm not sure how you'd like to approach that in the context of a method like `from_parsed_value()` where the `uri` is already given?

`ParsedValue` enum leveraged the serde untagged convenience to map the formats own Value type, but to have that done for `config::Value` instead just to add the `uri` would be verbose of an implementation?

I understand if using an intermediary this way instead of skipping that may seem as a codesmell, and I think I may have originally intended to look into how I could improve that after this PR was resolved. Partly because of a lack of familiarity with the project as a whole and not wanting to introduce breaking changes or deal with extra churn / considerations that may cause vs this reduced scope that was easier for me to reason about and get through review 😅 

As such, I still think the current approach with this PR using the intermediary enum would be valid and keeps the PR focused on the goal of unifying the serde compatible formats to a common parse method.

Refactoring that to then make `ParsedValue` redundant and work directly with `config::Value` + `ValueKind` seems better suited to a follow-up PR, and would be more pleasant in commit/merge history as context for these distinct changes to any interested party, as opposed to munged together (_but that could just be more perspective as I tend to encourage more focused PRs with all commits squashed via the PR merge strategy_).

---

All the `uri` value represents is the source, so for these formats the `config::Value` returned and any nested value produced, that `uri` is the same throughout.

As such it's not that meaningful during the format mapping to `config::Value`, but collections of `config::Value` need each item annotated with that same `uri` value/ref, so that merging/layering has that traceability of source.

If that were something better handled to avoid the need for `ValueKind::Array` and `ValueKind::Table` to require `config::Value` items just to track the associated `uri`, then this whole thing could be simplified as desired 😅

polarathene: > Building up my `serde-basepath` idea I've been toying with

I glanced through the linked ref but that still didn't click for me sorry.

---

In the current state of this PR, I could have `from_parsed_value()` assign `None` as the `uri` (_in lieu of a `serde-untagged` implementation effectively transforming to `Value::new(uri, kind)` for deserializing to `config::ValueKind` on the `Array`/`Table` variants, or [`kind.into()`](https://github.com/rust-cli/config-rs/blob/07f13ff25f18e6a7bd40a2d8a2f1e036a473736a/src/value.rs#L865-L875) if deserializing straight to `config::Value`?_), and then I could call a separate helper to update all the ~~`uri`~~ `origin` to the desired one before returning the final `config::Value` result?

```rust
// In value.rs:
impl Value {
    pub fn set_origin_recursively(&mut self, origin: Option<&String>) {
        self.origin = origin.cloned();
    
        // If kind is a collection, update the `origin` of their `Value`s too:
        match self.kind {
            ValueKind::Array(ref mut collection) => {
                for v in collection.iter_mut() {
                  v.set_origin_recursively(origin)
                }
            }
            
            ValueKind::Table(ref mut collection) => {
                for (_, v) in collection.iter_mut() {
                  v.set_origin_recursively(origin)
                }
            }
    
            // This is presumably unacceptable for handling all other variants
            // of `ValueKind` that don't contain nested `Value`s, but short of listing
            // all remaining variants here I'm not sure how you'd want to approach it
            _ => {},
        };
    }
}

// format.rs:
// Nothing to map if already deserialized into Value:
pub fn from_parsed_value(uri: Option<&String>, mut value: Value) -> Value {
    value.set_origin_recursively(uri);
    value
}

// Alternatively for ValueKind (implies Array/Table variants already have `v.origin = None`):
pub fn from_parsed_value(uri: Option<&String>, value: ValueKind) -> Value {
    let mut v = Value::from(value);
    v.set_origin_recursively(uri);

    v
}
```

Both would make `from_parsed_value()` largely redundant then 😅

The above `set_origin_recursively()` utility method could be added via a separate PR in the meantime? (_`origin` isn't public, so updating the `origin` after it's set isn't supported I think?_)

polarathene: I suppose with the more verbose addition of `serde-untagged`, handling the `Array` + `Table` types of `ValueKind` would be a non-issue as `.map`/`.seq` calls would could leverage `.into()` similar to how you could construct `ValueKind` like this:

```rust
// ValueKind::Table(T) storing `Vec<T> => ValueKind::Array(T) => Value`
let map_of_vecs: ValueKind = HashMap::<String, Value>::from([
    ("abc".to_string(), vec!["a", "b", "c"].into()),
    ("xyz".to_string(), vec!["x", "y", "z"].into()),
]).into();

// ValueKind::Table(T) => Value
let v = Value::from(map_of_vecs);
```

[`Value::from`](https://github.com/rust-cli/config-rs/blob/07f13ff25f18e6a7bd40a2d8a2f1e036a473736a/src/value.rs#L865-L875) will provide the `None` placeholder for `origin` and the added recursive method can patch that afterwards 👍

---

I think these two collection variants of `ValueKind` were only problematic with `serde(untagged)` when paired with the `from_parsed_value()` recursively calling itself to map the enum value, since they were not wrapping `Self` 😅

Other than that, there's the `Option` value type from Ron which I'm not sure how you'd approach that with `serde-untagged`, for `serde(untagged)` with `ParsedValue` the recursive call through `from_parsed_value()` [worked with the type `Option<Box<Self>>`](https://github.com/rust-cli/config-rs/pull/472/commits/e597f6aaef1205d508480735bf887bb8301408d9).

Likewise, while `serde-untagged` and formats like Ron support `char`, `ValueKind` lacks it and casts to `ValueKind::String`. A similar concern exists for Toml support when parsing it's date value as there is no equivalent in `ValueKind` it is mapped to the string variant too. Both cases were handled in this PR with a separate `ParsedString` enum delegated to `serde(untagged)`, all wrapped in a custom deserializer method for the `ParsedValue::String` variant and the `deserialize_with` annotation (_that I've not looked into if that's compatible with `serde-untagged`, I assume it is?_).

So a few gaps on how to proceed there with `serde-untagged`, but I think it might still be better deferred to a follow-up PR replacing `serde(untagged)`. It's diff might then be more straight-forward?

polarathene (commented):
Apologies for the verbosity, limited time to condense all my comments but I'll try 💪 

---

The gist of my response is as follows:

- `ParsedValue` + `from_parsed_value()` unifies the equivalent code it replaces. Deserializing directly to `config::Value` would be a separate improvement, this PR just consolidates / streamlines existing mapping. It might not be ideal, but I think it's still worthwhile?
- I don't think performance is the deciding factor for choosing `config-rs` over the competition. I'm open to adopting `serde-untagged` here if someone could fill the gaps in proceeding, but it might be better deferred to a separate PR at a later date?

---

## `ParsedValue` is not replacing `ValueKind`

> As this just replaces our existing type we are deserializing to, I'll give this a pass.

The PR DRYs up existing format implementations (_which are inconsistent in style/approach for blocks that accomplish the same task_).

Assuming that your comment was referring to the `ValueKind` type being replaced (_but I could be mistaken and you actually meant each individual format value type_):
- `ParsedValue` substitutes each formats own Value type, while `from_parsed_value()` unifies the existing equivalent format methods mapping of a format specific `Value` type to `config::ValueKind` (_and subsequent conversion to `config::Value`_).
- The `from_*_value()` format methods tend to be recursively called for mapping to the final `config::Value` type (_optionally with nested `config::Value` when `ValueKind` is a collection type_) and associating the fixed `origin` source to each mapped `config::Value`.
- `ValueKind` contains collection variants that store items of `config::Value` not `Self`, which is problematic for recursive calls of `from_*_value()`. You would instead need to deserialize directly into `config::Value` AFAIK and then recursively update `origin`.

This transition of existing formats to a common `ParsedValue` enum + `from_parsed_value()` method is AFAIK an improvement for maintainability and anyone referencing the format support to contribute new formats.

When I authored this PR I was able to easily add a few other formats without copy/pasting boilerplate of one of the existing implementations. That seems worthwhile?

---

## `serde-untagged` / performance

> `serde(untagged)` is a performance footgun in situations like this (easily noticeable in benchmarks when I made the same mistake in `toml::Value`).
> 
> We'd likely want to use `serde-untagged`.

Personally that may not be as digestible of a change within this PR?
- The `serde(untagged)` vs `serde-untagged` concern on performance is valid, but IMO is better served as a follow-up PR to this one if it will result in additions [akin to this support for `config::Value`](https://github.com/rust-cli/config-rs/blob/07f13ff25f18e6a7bd40a2d8a2f1e036a473736a/src/value.rs#L710-L863) (_and [this much more verbose related serde support](https://github.com/rust-cli/config-rs/blob/07f13ff25f18e6a7bd40a2d8a2f1e036a473736a/src/de.rs)_).
- No one has tackled the concern this PR addresses since I raised it, so I'm not sure if blocking on requiring additional complexity will be beneficial 😓

That said, should anyone find the performance an issue when parsing configs in their real-world projects, they'd likely be motivated to contribute that suggested improvement (_which is either reverting to the current state we have now prior to this PR with it's drawbacks, or adapting from the changes of this PR to use `serde-untagged` and skip `ParsedValue`?_), you're an example of that with your contributions to the toml crates, so I completely understand the reluctance 😅 

Config parsing benchmarks are presumably a rather niche concern for the bulk of projects though (_which might prefer a config crate that is easier to grok and maintain, as existing issues tend to be focused on_).

As you're effectively the core maintainer of this crate now, that is totally up to you. I am in support of your requests if you're familiar enough with how to approach the concerns I raised in my feedback that are unclear to me, but I get the impression it'd still be better delegated to a subsequent PR? (_it has already been [suggested that this PR alone in it's current state would benefit from splitting](https://github.com/rust-cli/config-rs/issues/590#issuecomment-2434001971)_)

src/format.rs
epage: Unless there is a significant problem we run into, I stand by using `serde-untagged`as I don't want the other formats to regress by adding this performance hit.

> I don't have a problem with that, but it was nice to have for simple mappings of types to their variant.

I tried to push for serde_derive to have attributes to force type-variant mappings to avoid the performance overhead of `untagged` but the maintainers weren't interested.

polarathene: > I stand by using `serde-untagged`as I don't want the other formats to regress by adding this performance hit.

That's fair.

It could be useful as an alternative format helper to implement additional formats, which although might not be used by any official format, if a user could extend `config-rs` support separately without requiring upstreaming of a format into the crate?

I'll look into adapting to `serde-untagged` when I can find the time, but if anyone else wants to tackle that it'd be appreciated 😅

---

> I tried to push for `serde_derive` to have attributes to force type-variant mappings to avoid the performance overhead of `untagged` but the maintainers weren't interested.

That's unfortunate, sounds like a very reasonable request 🤔

polarathene:
I'm detailing this as a refresher for future me or anyone else who may be interested in pushing this PR forward.

My time is often sparse, so with the additional requirements to adopt `serde-untagged` and the lack of clarity of what some `serde(untagged)` equivalent functionality would look like with `serde-untagged`, my availability to resume this PR is hindered.

**Remaining tasks:**
- [ ] Rebase and resolve conflicts.
- [ ] Introduce the [`set_origin_recursively()` method](https://github.com/rust-cli/config-rs/pull/472#discussion_r2264510051) to support deserializing straight into `config::Value` (_making redundant both `from_parsed_value()` from this PR, and the format equivalent methods it would have replaced_).
- [ ] Migrate from `serde(untagged)` to `serde-untagged` due to performance concerns.
  - Potentially blocked by lack of clarity on how to proceed with some variant handling/mismatch (_as [detailed here](https://github.com/rust-cli/config-rs/pull/472#discussion_r2264510307)_), such as Ron's [`Option(Option<Box<Value>>)`](https://github.com/ron-rs/ron/blob/29e244a21cc8eefeceb6a3a7e2be71760a0ce792/src/value/mod.rs#L28) and this PR's `ParsedString` enum that helps deserialize several formats values to `ParsedValue::String(String)`.
  - May still be useful to [continue deserializing into `ParsedValue` enum](https://github.com/rust-cli/config-rs/pull/472#discussion_r2264167658) instead of `Value` struct / `ValueKind` enum directly, given that [`ParsedValue` is independent of `origin`/`uri`](https://github.com/rust-cli/config-rs/pull/472#discussion_r2264173375).
- [ ] Ensure [concerns regarding Ron and YAML support](https://github.com/rust-cli/config-rs/issues/675#issuecomment-3166247688) have been addressed.
  - Switch YAML crate to `saphyr` (_[if it lands serde support](https://github.com/saphyr-rs/saphyr/pull/53)_).
  - Verify newer Ron lossless number types are converted to the [expected `ValueKind` types](https://github.com/rust-cli/config-rs/pull/476). There is a related support concern [regarding `u128` / `i128` with `config-rs`](https://github.com/rust-cli/config-rs/pull/476/files#r2003572772).
  - Ron 0.9 [introduces a `Value::Bytes(Vec<u8>)` variant](https://github.com/ron-rs/ron/pull/438) for representing byte strings unambiguously, presently with this PR that would deserialize into `ValueKind::Nil` as there is no supported mapping (_or [could be `ValueKind::Array` with `Vec<u8>`](https://github.com/rust-cli/config-rs/pull/472#issuecomment-1765609751)?_)
- [ ] Follow-up with PRs for [supporting additional formats](https://github.com/rust-cli/config-rs/pull/472#issuecomment-1770151034) (_`jsonc` support will likely have merged before this PR does, and [could likewise benefit as it has serde support](https://github.com/rust-cli/config-rs/pull/457#issuecomment-3173241589)_).

src/format.rs
epage: > It could be useful as an alternative format helper to implement additional formats, which although might not be used by any official format, if a user could extend config-rs support separately without requiring upstreaming of a format into the crate?

They can with https://docs.rs/config/latest/config/trait.Format.html

They won't get our [extension-agnostic loader](https://docs.rs/config/latest/src/config/file/source/file.rs.html#78-87) but that isn't always necessary and can easily be implemented on its own.

epage:
> Introduce the https://github.com/rust-cli/config-rs/pull/472#discussion_r2264510051 to support deserializing straight into config::Value (making redundant both from_parsed_value() from this PR, and the format equivalent methods it would have replaced).

Note that I said this was non-blocking but we will eventually want to do this.

epage:
> Ensure https://github.com/rust-cli/config-rs/issues/675#issuecomment-3166247688 have been addressed.

It is non-blocking to update all formats to this

epage:
Use of `serde_untagged` is blocked on dtolnay/serde-untagged#10 being merged

epage:
#677 was merged

polarathene:
Although I now have the time available, recent experiences have led me to shift my focus away from `config-rs` to other projects.

Feel free to close this PR or leave it open for visibility should someone else want to progress it.

orangecms:
To anyone still interested, there is now https://github.com/bourumir-wyngs/serde-saphyr (current focus, looks promising, I think) as well as 
https://github.com/bourumir-wyngs/serde-yaml-bw (looks more stable).

Related:
- https://github.com/saphyr-rs/saphyr/issues/66 (I think that's resolved now, was just a bit of a messy discussion)
- https://github.com/SergioBenitez/Figment/pull/154 (suggesting and discussing a few other options)

PR #474, opened 2023-10-10
fix: Change YAML crate to `serde_yaml`

polarathene:
- `yaml-rust` is unmaintained for over 2 years.
- Like with `yaml-rust`, requires special handling for table keys (_ensuring numbers are converted to strings_).

Resolves: https://github.com/mehcode/config-rs/issues/473

---

Just some insights below that might be worth noting considering the switch. Presumably this should be delayed until a breaking release is acceptable?

Since `config-rs` is only interested in deserialization, some [concerns users may have regarding serializing valid YAML files](https://github.com/dtolnay/serde-yaml/issues/380) may be a non-issue, but changing from one YAML crate to another may behave differently with deserialization still:
- Prior to the [`0.9` release](https://github.com/dtolnay/serde-yaml/releases/tag/0.9.0) (July 2022) of `serde_yaml`, it used the `yaml-rust` crate internally. Now it uses a [transpiled `libyaml` crate](https://github.com/dtolnay/unsafe-libyaml).
- The `yaml-rust` format had parsing logic by `config-rs` to ensure a single document or create an empty Map. You should get [similar with `serde_yaml`](https://github.com/dtolnay/serde-yaml/releases/tag/0.9.20) for an empty Map or `ValueKind::Nil`
- [Numbers can be keys](https://github.com/stoplightio/yaml/issues/34), but have always been converted to strings for the `Map` type in `config-rs`.
  - There is an existing [test-case YAML config for this](https://github.com/mehcode/config-rs/blob/55c464e143d7301f3c107a405827fe9dc085f81a/tests/test-keys.yaml) with mixed number and string keys. No difference in behaviour here than what we already had.
  - There is a [tracking issue at `serde_yaml`](https://github.com/dtolnay/serde-yaml/issues/165#issuecomment-644311452), ideally that would be resolved upstream to properly handle number to string conversion (_only seems to be an issue for untagged enums_).
- `serde_yaml` changed tagged enums with [`0.9` release](https://github.com/dtolnay/serde-yaml/releases/tag/0.9.0), that is [causing some issues from users upgrading from `0.8`](https://github.com/dtolnay/serde-yaml/issues/361). `config-rs` test suite is passing, but the test coverage for config files input to Rust types may not cover some of these concerns? 🤷‍♂️ 
- Some YAML parsers have a wider range of values that are treated as booleans based on YAML 1.1 spec apparently, while [`serde_yaml` only implements support for what the YAML 1.2 spec states should be treated as bools](https://github.com/dtolnay/serde-yaml/pull/330).
  - A [request for wider support was rejected](https://github.com/dtolnay/serde-yaml/issues/379#issuecomment-1621990180).
  - It [may be encountered in some configs like Ansible](https://github.com/dtolnay/serde-yaml/issues/384), where it's considered a valid bool value type. I've not checked if `yaml-rust` is any better at this though.
- What may look like [an invalid config, may be treated as a single `String`](https://github.com/dtolnay/serde-yaml/issues/372) instead of throwing an error depending on the input, as it's technically valid.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,7 +17,7 @@ maintenance = { status = "actively-developed" }
 [features]
 default = ["toml", "json", "yaml", "ini", "ron", "json5", "convert-case", "async"]
 json = ["serde_json"]
-yaml = ["yaml-rust"]
+yaml = ["serde_yaml"]
 ini = ["rust-ini"]
 json5 = ["json5_rs", "serde/derive"]
 convert-case = ["convert_case"]
@@ -32,7 +32,7 @@ nom = "7"
 async-trait = { version = "0.1.50", optional = true }
 toml = { version = "0.8", optional = true }
 serde_json = { version = "1.0.2", optional = true }
-yaml-rust = { version = "0.4", optional = true }
+serde_yaml  = { version = "0.9", optional = true }
 rust-ini = { version = "0.19", optional = true }
 ron = { version = "0.8", optional = true }
 json5_rs = { version = "0.4", optional = true, package = "json5" }
--- a/README.md
+++ b/README.md
@@ -18,7 +18,7 @@
 
 [JSON]: https://github.com/serde-rs/json
 [TOML]: https://github.com/toml-lang/toml
-[YAML]: https://github.com/chyh1990/yaml-rust
+[YAML]: https://github.com/dtolnay/serde-yaml
 [INI]: https://github.com/zonyitoo/rust-ini
 [RON]: https://github.com/ron-rs/ron
 [JSON5]: https://github.com/callum-oakley/json5-rs
--- a/src/file/format/mod.rs
+++ b/src/file/format/mod.rs
@@ -40,7 +40,7 @@ pub enum FileFormat {
     #[cfg(feature = "json")]
     Json,
 
-    /// YAML (parsed with yaml_rust)
+    /// YAML (parsed with serde_yaml)
     #[cfg(feature = "yaml")]
     Yaml,
 
--- a/src/file/format/yaml.rs
+++ b/src/file/format/yaml.rs
@@ -1,8 +1,4 @@
 use std::error::Error;
-use std::fmt;
-use std::mem;
-
-use yaml_rust as yaml;
 
 use crate::format;
 use crate::map::Map;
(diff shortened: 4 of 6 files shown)


polarathene:
Note, I have implemented the equivalent for https://github.com/mehcode/config-rs/pull/472 which could include the `serde_yaml` support switch instead if preferred (_or addressed as a separate follow-up PR_).

Presently the [`Tagged` variant of `serde_yaml::Value`](https://github.com/dtolnay/serde-yaml/blob/25f218b8731f2a1c8e01921408e67d2f21cd083c/src/value/mod.rs#L26-L43) is not handled. I don't think we can easily implement support for that conversion either? I'm not experienced enough at least to know how to integrate that.

The feature has `!Variant` as a value prefix in a `.yaml` file, and `serde_yaml` represents the type with two properties, `tag` (_that `!Variant` value_) and `value` actual data for the enum variant.

I'm inclined to ignore it since it's not a previously supported feature 🤷‍♂️

matthiasbeyer:
First of all, sorry for the delay. I got sick and am writing this from my bed.

So what I see here is a lot of change in behavior when switching from one yaml implementation to another. I am not sure whether I like that, but on the other hand we are in a zero-dot version (`0.x.y`), so it wouldn't be an issue from a version perspective.

I'll leave it up to you, really!

polarathene:
> I got sick and am writing this from my bed.

Ah, no worries rest up 👍 

I pinged due to your activity appearing active and wasn't sure if you were notified of the PR.

---

> So what I see here is a lot of change in behavior when switching from one yaml implementation to another.

It's roughly at parity, at least the implementation is very similar on `config-rs` end and what is supported.

I've otherwise made note above of known issues. Some of these may not be relevant to `config-rs` as it deserializes from YAML input to `Value` as an intermediate / normalized storage before deserializing to a struct provided by the user? This adds another layer of deserialization where types may be converted to the appropriate type to satisfy the users config struct fields.

> on the other hand we are in a zero-dot version (`0.x.y`), so it wouldn't be an issue from a version perspective.

Tests were helpful and ensured that the switch at least passes those, if there is something that wasn't covered, new tests could be added. It may warrant bumping to a `0.14` release? 🤷‍♂️ (_or a part of that bump_)

> I'll leave it up to you, really!

My preference is to #472 which already bundles the equivalent feature change, but without the `serde-yaml` specific enum handling.

The yaml format file becomes very simple with one special case for the yaml maps (_due to string keys for table being necessary_).

There may be some differences between this PR and #472 untagged enum approach, but current tests pass all the same. In the event there is an issue, this PR provides an alternative to fallback to.

matthiasbeyer:
Note to self:

> My preference is to https://github.com/mehcode/config-rs/pull/472 which already bundles the equivalent feature change, but without the serde-yaml specific enum handling.

polarathene:
Now that the MSRV lock file issue is out of the way, I'll be revisiting my PRs this weekend. If you want to block on any let me know, otherwise I'll self-review again and merge after addressing any existing concerns I raised.

matthiasbeyer:
> I'll self-review again and merge after addressing any existing concerns I raised.

Please do that! Thank you so much for your contributions!

epage:
`serde_yaml` is deprecated.

We now use `yaml_rust2`

PR #480, opened 2023-10-23
Hide and deprecate AsyncConfigBuilder

ijackson:
This seems to be a nonfunctional vestige.


ijackson:
> gitlint will also complain here about long lines in the commit. Do you want me to take over to safe you the hassle?

Does that mean I'm not to quote error messages in commit messages?  That's a shame.  Yes, please, do feel free to take this over.

PR #488, opened 2023-10-23
Unify deser impl (redux)

ijackson:
This is #338 reworked to apply to current main, and with the test case simplified as requested.

### DIFF
--- a/src/de.rs
+++ b/src/de.rs
@@ -363,122 +363,56 @@ impl<'de> de::VariantAccess<'de> for EnumAccess {
     }
 }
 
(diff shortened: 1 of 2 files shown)


matthiasbeyer (approved):

polarathene:
I haven't checked yet, but heads-up prior to all the rework done recently to the `ser.rs` and `de.rs`, I had identified the cause of https://github.com/mehcode/config-rs/issues/461#issuecomment-1771963758 and fix. Although that's less obvious now.

Might be worth verifying if you've implicitly resolved that issue too, or added more drift 😅

PR #490, opened 2023-10-23
deserialize: strings: Introduce Unreachable type

ijackson:
This allows us to remove a lot of code that contains calls to unreachable! (ie, a runtime panic) with code that demonstrates, in a way that's checked by the compiler, that the call cannot happen.


matthiasbeyer:
I rebased this and will merge if CI is green.

PR #493, opened 2023-10-26
CI speed optimization

### LINKED ISSUE #482: Investigate ways to make CI faster
I really don't like how slow CI is for this repository.

If someone wants to make CI faster, I'd love to see PRs for that!

matthiasbeyer:
This PR experiments with CI speedup optimizations.

I didn't realize that CI in this project did not even use caching... :eyes: 

---

Closes #482 
... probably? :eyes:

### DIFF
--- a/.github/workflows/msrv.yml
+++ b/.github/workflows/msrv.yml
@@ -60,7 +60,7 @@ jobs:
     strategy:
       matrix:
         rust:
-          - stable
+          - 1.70.0
           - beta
     steps:
       - name: Checkout sources
--- a/.github/workflows/msrv.yml
+++ b/.github/workflows/msrv.yml
@@ -12,7 +12,6 @@ jobs:
           - 1.70.0
           - stable
           - beta
-          - nightly
 
     steps:
       - name: Checkout sources
@@ -27,12 +26,6 @@ jobs:
           toolchain: ${{ matrix.rust }}
 
       - name: Run cargo check
-        if: matrix.rust != 'nightly'
-        run: cargo check --all-features
-
-      - name: Run cargo check (nightly)
-        if: matrix.rust == 'nightly'
-        continue-on-error: true
         run: cargo check --all-features
 
   test:
@@ -45,7 +38,6 @@ jobs:
           - 1.70.0
           - stable
           - beta
-          - nightly
     steps:
       - name: Checkout sources
         uses: actions/checkout@v4.1.1
@@ -59,17 +51,6 @@ jobs:
           toolchain: ${{ matrix.rust }}
 
       - name: Run cargo test
-        if: matrix.rust != 'nightly' && matrix.rust != '1.56.1'
-        run: cargo test --all-features
-
-      - name: Run cargo test (nightly)
-        if: matrix.rust == '1.66.0'
-        continue-on-error: true
-        run: cargo test --tests --all-features
-
-      - name: Run cargo test (nightly)
-        if: matrix.rust == 'nightly'
-        continue-on-error: true
         run: cargo test --all-features
 
   fmt:
--- a/.github/workflows/msrv.yml
+++ b/.github/workflows/msrv.yml
@@ -43,14 +43,17 @@ jobs:
       - name: Cache
         uses: Swatinem/rust-cache@v2
 
+      - name: Install nextest
+        uses: taiki-e/install-action@nextest
+
       - name: Install toolchain
         uses: dtolnay/rust-toolchain@master
         with:
           toolchain: ${{ matrix.rust }}
 
-      - name: Run cargo test
+      - name: Run cargo nextest
         continue-on-error: ${{ matrix.rust == 'beta' }}
-        run: cargo test --all-features
+        run: cargo nextest run --all-features
 
   fmt:
     needs: [check]
--- a/.github/workflows/msrv.yml
+++ b/.github/workflows/msrv.yml
@@ -26,6 +26,7 @@ jobs:
           toolchain: ${{ matrix.rust }}
 
       - name: Run cargo check
+        continue-on-error: ${{ matrix.rust == 'beta' }}
         run: cargo check --all-features
 
   test:
@@ -51,6 +52,7 @@ jobs:
           toolchain: ${{ matrix.rust }}
 
       - name: Run cargo test
+        continue-on-error: ${{ matrix.rust == 'beta' }}
         run: cargo test --all-features
 
   fmt:
@@ -73,6 +75,7 @@ jobs:
           components: rustfmt
 
       - name: Run cargo fmt
+        continue-on-error: ${{ matrix.rust == 'beta' }}
         run: cargo fmt --all -- --check
 
   clippy:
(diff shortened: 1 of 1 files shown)


matthiasbeyer:
Some more changes, lets see how fast we're now. :laughing:

matthiasbeyer:
Ok, I got 30 secs faster CI now... that's not nearly enough IMO. Could even be just random noise.

matthiasbeyer:
I wonder why the tests for 1.70.0 is so much slower than the ones for 1.73.0 ... nextest did not actually make anything faster. Probably because building is slow, not running. But why is the job for 1.70.0 running 2:20min, and the 1.73.0 one only 58s? I don't know.

So overall we're now slower than before. :cry: 

Let's see what happens if we execute tests right away and not wait for the check phase.

matthiasbeyer:
4 min -> 3 min now.

Not sure whether this is actually worth it... but I guess I am ready for review here. Just a quick squash of the fixup commit.

polarathene:
> I wonder why the tests for 1.70.0 is so much slower than the ones for 1.73.0

`1.73.0` being the latest stable right? I already mentioned that Github likely caches this themselves as they're known to do with other high traffic sources for CI. Far cheaper for them to pull from the real location once every N (eg: minute), than every CI job running within N duration on their infrastructure.

Older toolchains are less likely to have that optimization? 🤷‍♂️ (_just an assumption, that there is something different in the requests that differentiates this_)

I believe it's a common practice, even for ISPs to cache popular content from vendors like Youtube / Netflix (_pretty sure I had read of partnerships to do such_).

---

> nextest did not actually make anything faster.

I also mentioned that observation, tests did not seem to take much of the time so I didn't see it providing much benefit.

> So overall we're now slower than before

Keep in mind that when using `actions/cache` action, it will have access to `main` branch cache (once CI runs on that), and then PRs can use that. However PRs maintain their own scoped cache IIRC, and that is not accessible to another PR, nor `main` I think 🤷‍♂️ 

Should be in the docs, I could be recalling that wrong. Caching makes a big difference for `docker-mailserver` builds.

polarathene:
[As mentioned](https://github.com/marketplace/actions/rust-cache#cache-effectiveness):

image

Additionally keep in mind what is being cached, and the cache key generation:

image

This also confirms that they're using the official github cache action, just extending it to better suite cargo:

image

---

We can see that the action is correctly saving and restoring cache quickly, but pay attention to this:

image

Your current PR is using the action, then setting a toolchain, which is against the actions advice.

.github/workflows/msrv.yml
```diff
@@ -10,61 +10,49 @@ jobs:
...
-          - stable
-          - beta
-          - nightly
 
     steps:
       - name: Checkout sources
         uses: actions/checkout@v4.1.1
 
+      - name: Cache
+        uses: Swatinem/rust-cache@v2
```
polarathene, line 19: Move this after you have installed your toolchain.

If you are only going to use the projects MSRV, you're not using a matrix anymore for this job and could adjust that.

.github/workflows/msrv.yml
```diff
@@ -10,61 +10,49 @@ jobs:
...
   test:
-    needs: [check]
     name: Test Suite
     runs-on: ubuntu-latest
     strategy:
       matrix:
         rust:
           - 1.70.0
-          - stable
+          - 1.73.0 # stable
```
polarathene, line 36: Probably should keep that as `stable`?

.github/workflows/msrv.yml
```diff
@@ -10,61 +10,49 @@ jobs:
...
-          - nightly
     steps:
       - name: Checkout sources
         uses: actions/checkout@v4.1.1
 
+      - name: Cache
+        uses: Swatinem/rust-cache@v2
+
+      - name: Install nextest
+        uses: taiki-e/install-action@nextest
```
polarathene, line 46: I'd just focus on the cache action in this PR, only bother with this if it shows any actual improvement. I don't think the `config-rs` test suite is large enough to benefit from that, as we've observed the slowdown was actually from building the crate.

.github/workflows/msrv.yml
```diff
@@ -10,61 +10,49 @@ jobs:
...
-          - stable
+          - 1.73.0 # stable
           - beta
-          - nightly
     steps:
       - name: Checkout sources
         uses: actions/checkout@v4.1.1
 
+      - name: Cache
+        uses: Swatinem/rust-cache@v2
```
polarathene, line 43: Move to after toolchain step.

.github/workflows/msrv.yml
```diff
@@ -86,6 +73,7 @@ jobs:
           components: rustfmt
 
       - name: Run cargo fmt
+        continue-on-error: ${{ matrix.rust == 'beta' }}
```
polarathene, line 76: This has no value if you're removing the `stable` and `beta` toolchains. Like mentioned earlier, if you plan to test only MSRV, then drop the matrix usage?

.github/workflows/msrv.yml
```diff
@@ -95,12 +83,19 @@ jobs:
...
+          # version. So there's no point in running it with two or more
+          # versions, MSRV should be fine.
           - 1.70.0
-          - 1.73.0
     steps:
       - name: Checkout sources
         uses: actions/checkout@v4.1.1
 
+      - name: Cache
+        uses: Swatinem/rust-cache@v2
```
polarathene, line 97: Move below toolchain step

polarathene (changes requested):
Just for clarification, you define multiple separate jobs to run, these are all via separate runners just like matrix AFAIK?

Each separate runner then is building the project again. You could instead try minimize the jobs?

---

Here is an example of managing a check status, which I think can be used to display multiple from the same job if that suits you:

- [Create new pending state](https://github.com/docker-mailserver/docker-mailserver/blob/097dc6c9a4b48d5955c4d9c40c191bfb4e0ca29d/.github/workflows/docs-preview-deploy.yml#L48-L56)
- [Update with success/failure status](https://github.com/docker-mailserver/docker-mailserver/blob/097dc6c9a4b48d5955c4d9c40c191bfb4e0ca29d/.github/workflows/docs-preview-deploy.yml#L108-L120)

This makes it visible:

image

Bit more verbose workflow, but could help.

---

Alternatively, as suggested before, have one job start per toolchain, then subsequent jobs [express a dependency](https://github.com/docker-mailserver/docker-mailserver/blob/097dc6c9a4b48d5955c4d9c40c191bfb4e0ca29d/.github/workflows/default_on_push.yml#L35) to the initial job and they could bring in that cache.

If you have multiple jobs run in parallel, then initially none have that cache and all need to do the same work.

Once you have got some cache, depending on how the cache key is defined, you can find jobs falling back to earlier cache they can use.

---

As per prior comment with images from the `rust-cache` docs, you may need to wait until the `Cargo.lock` is committed for the most benefit in cache usage.

.github/workflows/msrv.yml
matthiasbeyer: No, having a mutable pointer to the current stable broke our CI before because new lints were introduced in stable rust that failed for our codebase. I'd rather have a non-mutable version here and update it manually every now and then.

PR #495, opened 2023-11-05
[release-0.13.x] chore(Cargo.toml): Better document direct deps

polarathene:
These are improvements [from an extensive investigation](https://github.com/mehcode/config-rs/pull/492#issuecomment-1793319804) I recently did to simplify maintenance of lockfile generation for the CI to benefit from, while respecting the MSRV.

- Feel free to reject, although I don't see any downsides to merging this.
- Compliments: https://github.com/mehcode/config-rs/pull/492

**Commit message:**

<details>
<summary>Click to view</summary>

- Slight adjustments to the version fields for compatibility with `cargo +nightly update -Z direct-minimal-versions` with the MSRV of `1.56.0`.
- Add `rust-version` field for leveraging `cargo +nightly update -Z msrv-policy` to generate a lockfile that respects the MSRV, and the benefit of downstreams.
- Better communicate why `dev-dependencies` are required (_identify associated examples / tests_).
- Avoid repeating deps again in in `dev-dependencies`.
- Raise fixed `warp` dev dep to an MSRV compatible version (_for a common `tokio-util` implicit dep version_). Simplifies CI lock maintenance.

</details>

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,6 +10,7 @@ authors = ["Ryan Leckey <leckey.ryan@gmail.com>"]
 categories = ["config"]
 license = "MIT/Apache-2.0"
 edition = "2018"
+rust-version = "1.56.0"
 
 [badges]
 maintenance = { status = "actively-developed" }
@@ -23,13 +24,13 @@ json5 = ["json5_rs"]
 preserve_order = ["indexmap", "toml/preserve_order", "serde_json/preserve_order", "ron/indexmap"]
 
 [dependencies]
-async-trait = "0.1.50"
-lazy_static = "1.0"
-serde = "1.0.8"
+async-trait = "0.1.2"
+lazy_static = "1.4"
+serde = "1.0.99"
 nom = "7"
 
 toml = { version = "0.5", optional = true }
-serde_json = { version = "1.0.2", optional = true }
+serde_json = { version = "1.0.0", optional = true }
 yaml-rust = { version = "0.4", optional = true }
 rust-ini = { version = "0.18", optional = true }
 ron = { version = "0.7", optional = true }
@@ -38,16 +39,24 @@ indexmap = { version = "1.7.0", features = ["serde-1"], optional = true}
 pathdiff = "0.2"
 
 [dev-dependencies]
-serde_derive = "1.0.8"
 float-cmp = "0.9"
-chrono = { version = "0.4", features = ["serde"] }
-tokio = { version = "1", features = ["rt-multi-thread", "macros", "fs", "io-util", "time"]}
-warp = "=0.3.1"
-futures = "0.3.15"
-reqwest = "0.11.10"
 
-serde = "1.0"
+# tests/legacy (available in serde via `features = ["derive"]` since 1.0.103)
+serde_derive = "1.0.99"
+
+# tests/datetime.rs + tests/legacy/datetime.rs
+chrono = { version = "0.4.23", features = ["serde"] }
+
+# examples/async_source
+warp = "=0.3.5"
+futures = "0.3"
+reqwest = "0.11"
+# Additionally tests/async_builder.rs
+tokio = { version = "1.13", features = ["rt-multi-thread", "macros"]}
+
+# examples/glob
 glob = "0.3"
-lazy_static = "1"
-notify = "^4.0.0"
+# examples/watch
+notify = "4.0.0"
+# tests/env.rs
 temp-env = "0.2.0"
--- a/.github/workflows/msrv.yml
+++ b/.github/workflows/msrv.yml
@@ -9,7 +9,7 @@ jobs:
     strategy:
       matrix:
         rust:
-          - 1.59.0
+          - 1.56.1
           - stable
           - beta
           - nightly
@@ -48,7 +48,7 @@ jobs:
     strategy:
       matrix:
         rust:
-          - 1.59.0
+          - 1.56.1
           - stable
           - beta
           - nightly
@@ -67,14 +67,14 @@ jobs:
         run: cp Cargo.lock.msrv Cargo.lock
 
       - name: Run cargo test
-        if: matrix.rust != 'nightly' && matrix.rust != '1.59.0'
+        if: matrix.rust != 'nightly' && matrix.rust != '1.56.1'
         uses: actions-rs/cargo@v1
         with:
           command: test
           args: --all-features
 
       - name: Run cargo test (nightly)
-        if: matrix.rust == '1.59.0'
+        if: matrix.rust == '1.56.1'
         continue-on-error: true
         uses: actions-rs/cargo@v1
         with:
@@ -127,7 +127,7 @@ jobs:
       - name: Install toolchain
         uses: actions-rs/toolchain@v1
         with:
-          toolchain: 1.59.0
+          toolchain: 1.56.1
           override: true
           components: clippy
 
@@ -147,7 +147,7 @@ jobs:
     strategy:
       matrix:
         rust:
-          - 1.59.0
+          - 1.56.1
           - stable
 
     steps:
--- a/Cargo.lock.msrv
+++ b/Cargo.lock.msrv
@@ -96,15 +96,6 @@ checksum = "327762f6e5a765692301e5bb513e0d9fef63be86bbc14528052b1cd3e6f03e07"
 
 [[package]]
 name = "block-buffer"
-version = "0.9.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "4152116fd6e9dadb291ae18fc1ec3575ed6d84c29642d97890f4b4a3417297e4"
-dependencies = [
- "generic-array",
-]
-
-[[package]]
-name = "block-buffer"
 version = "0.10.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "3078c7629b62d3f0439517fa394996acacc5cbc91c5a20d8c658e77abd503a71"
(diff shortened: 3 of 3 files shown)


polarathene:
# More info

My verbosity is an issue for some, I've split this off and tried to condense / organize it. It's purely for reference to support and document the changes in the PR should anyone need more context.

## Producing a compatible `Cargo.lock` for the MSRV

```bash
# Resolve a `Cargo.lock` with the latest compatible deps for `rust-version` (1.56.0):
cargo +nightly update -Z msrv-policy

# Minor corrections due to deps lack of proper `-Z msrv-policy` support for the MSRV
cargo update --package memchr --precise 2.5.0
cargo update --package ordered-multimap --precise 0.4.0
cargo update --package tokio-util --precise 0.7.8
```

<details>
<summary>Validating minimum versions build for MSRV (with 1 less pin)</summary>

```
# For verifying minimum versions of direct deps build for the MSRV is supported:
# `tokio ="1.27"` (or lower) implicitly prevents needing to pin tokio-util to 0.7.8 (as newer releases require a newer tokio)
cargo +nightly update -Z direct-minimal-versions -Z msrv-policy
cargo update --package memchr --precise 2.5.0
cargo update --package ordered-multimap --precise 0.4.0
```

</details>

### Supplementary info on `-Z msrv-policy` and pinned versions

<details>
<summary>Click to view</summary>

`-Z msrv-policy` only works on nightly, but ideally resolves within the deps semver range a version that was last known as compatible with `rust-version` in `Cargo.toml`.
- This differs from the default resolver which will attempt to satisfy the latest version in semver regardless of `rust-version` compatibility.
- This also differs for older toolchains that don't understand newer `Cargo.toml` syntax like introduced in Rust `1.60.0`.
  - Those would ignore such published crates and resolve to an earlier version of the crate implicitly. These pins are the same versions that would be resolved regardless for the MSRV.
  - [`memchr 2.6.0`](https://github.com/BurntSushi/memchr/blob/c4bd85f58b763080fff52fbe67081decb2899b33/Cargo.toml#L40) is an implicit dep to quite a few crates in `Cargo.lock`, including the [explicit `nom 7` dep](https://github.com/rust-bakery/nom/blob/9678b061d3ed0a6b8e31ce569e024ec25f3dd354/Cargo.toml#L39). `memchr` only adopted [`rust-version` (1.60) since the `2.6.1` release](https://crates.io/crates/memchr/versions), hence `-Z msrv-policy` is ineffective at resolving without a pin, while the MSRV toolchain resolver would resolve a compatible version as anything newer is considered invalid.
  - `ordered-multimap` from `rust-ini` has the same problem, except it's history is [more complicated and documented in detail in this comment](https://github.com/mehcode/config-rs/pull/492#issuecomment-1793319804) (_[`ahash` is the issue](https://github.com/tkaitchuck/aHash/pull/181#discussion_r1377129069)_).
  - `tokio-util` is probably better pinned, that is easier by raising the explicitly pinned `warp` patch version in `Cargo.toml`. [This is also a bit complicated](https://github.com/mehcode/config-rs/pull/492#issuecomment-1793319968), but is an example of when `rust-version` is not correctly maintained.

</details>

<details>
<summary>Additional context on deps</summary>

**Minimum version:**
- `async-trait`: Minimum of `0.1.2` required to support `examples/async_source`.

**MSRV bumps:**
- `toml`: Newer releases after `0.5.x` series require MSRV 1.60 to 1.66, while the latest `0.8.6` release requires MSRV 1.67.
- `ron`: MSRV raised to 1.64 with `0.8.1`.
- `rust-ini`: MSRV raised to 1.64 with `0.19.0`, while `0.20.0` (latest) raises again to MSRV 1.71.1. Both due to `ordered-multimap`.
- `indexmap`: Next major `2.0.0` raises MSRV to 1.63.
- `notify`: The next major (`5.0.0`) keeps the 1.56 MSRV, while the latest `6.0.0` major raises that to MSRV 1.60
- `temp-env`: Newer minor release adopts `rust-version` since `0.3.4` (latest `0.3.6` with MSRV 1.62.1).

</details>

### Without `-Z msrv-policy`

The following will also support building without generating `Cargo.lock` with `-Z msrv-policy`. The versions need to be pinned (_in either `Cargo.toml` or `Cargo.lock`_).

<details>
<summary>Click to view</summary>

```toml
# These are the dev-dependencies with versions that last supported MSRV `1.56`

# Direct deps
tokio = { version = "1.29.1", features = ["rt-multi-thread", "macros"] } # latest 1.33 requires MSRV 1.63
chrono = { version = "0.4.26", features = ["serde"] } # newer versions raise MSRV to 1.57
reqwest = "0.11.15" # newer release bumps MSRV to 1.57 (latest 0.11.22 with MSRV 1.63)

# Implicit deps:
scoped-tls = "1.0.0" # 1.0.1 (latest) adds rust-version 1.59
byteorder = "1.4.3" # 1.5.0 (latest) adds rust-version 1.60 
log = "0.4.18" # newer versions add rust-version 1.60 (latest 0.4.20)
tempfile = "3.6.0" # newer versions raise to MSRV 1.63 (latest 3.8.1)
h2 = "0.3.20" # newer versions raise to MSRV 1.63 (latest 0.3.21)
tokio-util = "0.7.8" # newer versions incorrectly claim MSRV 1.56 (latest 0.7.10 with MSRV 1.63)
```

Alternatively pinning via `Cargo.toml` could avoid `=` pinning and instead rely on using `-Z direct-minimal-versions`. This requires extra pins due to the nightly toolchain resolver.

```toml
# Required only for nightly toolchain (eg: when using `-Z direct-minimal-versions`)
# For nom:
memchr = "2.5.0" # newer versions raise MSRV, but only set rust-version at 2.6.1, (latest 2.6.4 with MSRV 1.61)
# For rust-ini:
ordered-multimap = "0.4.0" # newer versions raise MSRV to 1.60 (latest 0.4.3 -> ahash 0.7.7)
# For json5:
# 2.5.7 for both pest deps is the last version supporting MSRV 1.56 (latest 2.75 with MSRV 1.61)
# (Avoidable with `-Z msrv-policy)
pest = "2.0.2"
pest_meta = "2.0.3"
```

The MSRV toolchain will resolve `memchr` and `ordered-multimap` correctly, thus they don't need to be pinned.
- Obviously the above isn't any better, but is shared to contrast against the `-Z msrv-policy` approach.
- 👎 Implicit deps (_even when only for `dev-dependencies` / CI usage_) should avoid being in `Cargo.toml`.
- I've documented the latest MSRV for the semver range of each dep, which better illustrates the motivation of [this recent PR](https://github.com/mehcode/config-rs/pull/487#issuecomment-1793326893) (_raising MSRV to `1.63` to keep the CI happy_).

</details>

---

## CI uses MSRV `1.59`, while this PR supports MSRV `1.56`

<details>
<summary>Click to view</summary>

Note that presently the CI is testing for an MSRV of `1.59`, while the README (for this branch) presently documents `1.56` as the MSRV.

- It was presumably bumped in the past to make the CI pass (_seems to be a common pattern_) without resolving a compatible lockfile.
- This PR enables maintaining the original MSRV of `1.56.0` (_or if preferred newer_).
  - `rust-version` is set to communicate the actual minimum that `config-rs` will build with.
  - `hashbrown 0.14.2` has done similar. Their `rust-version` is only supported if you additionally [pin a dep](https://github.com/rust-lang/hashbrown/blob/778e235eccc233f2d8b906ae8e4c3d717c44fcce/ci/run.sh#L19-L21), or opt-out of a default feature that enables it.

</details>

Cargo.toml
```diff
@@ -10,6 +10,7 @@ authors = ["Ryan Leckey <leckey.ryan@gmail.com>"]
 categories = ["config"]
 license = "MIT/Apache-2.0"
 edition = "2018"
+rust-version = "1.56.0"
```
polarathene, line 13: As noted by https://github.com/mehcode/config-rs/pull/483#issuecomment-1775534431
- This should only reflect what the project can build successfully with (_even if that requires [managing a `Cargo.lock` with some pinning](https://github.com/mehcode/config-rs/pull/495#issuecomment-1793634940)_).
- It should only be raised in future for explicit changes to the project (_not from implicit changes from deps semver raising MSRV_):
  - Adding/Upgrading deps in `Cargo.toml` where pinning cannot maintain support for `rust-version`.
  - Introducing changes that require features from newer toolchains.

When it comes to actual support / maintenance the MSRV policy of the project may differ. Raising `rust-version` is a more granular `edition` field, but with support to opt-out / ignore it by downstream.

ijackson:
I'm going to try to limit myself to comments which will hopefully allow us to move forward.

@polarathene Thanks for the alternative MR #495.

Thanks also for the investigations.  I agree that it's likely that the actual MSRV is 1.56 rather than 1.59.  I didn't set out, with this MR, to lower the MSRV of the 0.13.x branch.  But that seems like it would be a good thing to do.

I do think that we should continue to run tests and examples with our MSRV in CI.  I think we must do that with a fixed `Cargo.lock`, since otherwise our CI is likely to (once again) break due to changes in our dependency stack.

It is not so easy to produce a working `Cargo.lock`.  I note that #495 does not produce a `Cargo.lock` capable of running the tests and examples.  I used the runes in https://github.com/mehcode/config-rs/pull/495#issuecomment-1793634940, with that branch, and it didn't work for me.  (And, CI on that MR is failing, but that's partly due to that branch not having a `Cargo.lock` at all.)

My approach to producing a working `Cargo.lock` was indeed not very principled.  My own commit message was not intended as a recipe for future use.  It was intended as an explanation of what I had done, for traceability and documentation.

I am very open to an alternative method of generating a working `Cargo.lock`.  In whatever way we produce it, the method that was used generating it ought to be written down somewhere.  (And as I say I think it needs to be committed to the tree and used for our MSRV CI.)

If we can make a stable and reproducible way of generating a working `Cargo.lock` that would be ideal; we could even commit the generation script.  That's [what we do](https://gitlab.torproject.org/tpo/core/arti/-/blob/main/maint/downgrade_dependencies?ref_type=heads) in the Arti project.  But my attempts to do that for config-rs were not successful.

The concern about dependabot could perhaps be resolved by committing the `Cargo.lock` under a different name.  In [derive-adhoc](https://lib.rs/crates/derive-adhoc) I have a [Cargo.lock.minimal](https://gitlab.torproject.org/Diziet/rust-derive-adhoc/-/blob/main/Cargo.lock.minimal?ref_type=heads).  We would have to change the CI to `cp Cargo.lock.minimal Cargo.lock` in the MSRV tests.

polarathene:
Full reproduction environment to verify generated lockfile:
https://github.com/mehcode/config-rs/pull/492#issuecomment-1797948314

Rebased onto #496 as requested. CI passes 👍 

---

I've additionally:
- Updated the `Cargo.lock.msrv` to reflect `Cargo.toml` changes, which also verifies the simpler lock file generation.
- Updated the CI to relax the MSRV tested back to `1.56.1` (_README states `1.56.0`, which `rust-version` matches_).

All tests pass again 👍 

**Awaiting approval** from either @ijackson or @matthiasbeyer

ijackson:
Thanks, excellent work.  It's great that we now have a non-awful method for generating a working lockfile.  I have reviewed the changes and verified the lockfile creation.

The MSRV reduction is also nice.

Please do merge this.

matthiasbeyer:
I am on vacation until late tomorrow. I hope I can get to having a look on my rather long backlog for this repository after tomorrow! Please _do_ bug me if I don't!

matthiasbeyer (approved):

PR #500, opened 2023-11-22
Release 0.13.4 changelog

### LINKED ISSUE #499: Requesting a release from the 0.13 branch
Hi.  Might it be possible to do a 0.13.x point release?

In the Arti project we are blocked on #464 (our ticket https://gitlab.torproject.org/tpo/core/arti/-/issues/1058) which was fixed in config-rs mainline in #465, backported to 0.13.x in #486.

If we aren't able to update to a config-rs with the fix for #464 we may need to temporarily fork the crate on crates.io, which would be quite undesirable.  Our workaround for #464 has a poor impact on the config format we accept, so our internal target for removing that workaround, that we set about a month ago, was to fix this for our own next release, which is due at roughly the end of November.

So ideally we'd be able to pick up a new config-rs 0.13.x in the next week or so.

Thanks for your attention.

matthiasbeyer:
Closes #499

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,26 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## Unreleased
 
+## 0.13.4 - 2023-11-22
+
+- Relaxed the MSRV to 1.56.0 for this release. Please have a look at
+  [#495] for details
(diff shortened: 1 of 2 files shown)


matthiasbeyer:
Running this through CI and if its green we will have a new release shortly.

PR #506, opened 2023-12-15
Add try parse ini

mfirhas:
Nested configs from .ini file cannot be converted automatically into struct's field types.

Add parsing for nested .ini file format.


polarathene (changes requested):
This PR is carrying over changes from the `0.13.x` branch unrelated to the actual PR intent 👎 

The related issue https://github.com/mehcode/config-rs/issues/504#issuecomment-1857147190 has been identified as not something that should try to be fixed in `file/format/ini.rs`, as a fix for one problem would introduce a regression that can affect others.

PR should **not** be merged.

matthiasbeyer:
> PR should not be merged.

Agreed.

mfirhas:
@polarathene Sorry for the inconvenient. I branched out from 0.13.4 in my fork, because that's work for me in my test like I explained in the issue. I though all the commits in release tags should be in master(I assume master is the most stable), turned out it should be cherry-picked.

PR #507, opened 2023-12-15
Check external types

### LINKED ISSUE #505: Add a `cargo check-external-types` lint in CI
`cargo check-external-types` should be added to our CI setup to verify what types are exposed from the `config` API.

This way we can get more confidence when to update dependencies and how we have to release those updates.

matthiasbeyer:
Closes #505


matthiasbeyer:
Looks good I guess.

PR #510, opened 2024-02-01
Prepare release 0.14.x

### LINKED ISSUE #509: New release/sync from master
Yesterday I ran into an issue where environment variable overrides weren't working correctly and I realized it's because of [this line](https://github.com/mehcode/config-rs/blob/a38387955622e6b8d48a49352629cd329ad6dc29/src/de.rs#L249) which does a case-sensitive comparison when looking for a variant constructor, but even changing the capitalization of the environment variable doesn't fix it because they get lower-cased in `env.rs`.  I notice that [master has a fix for this to be case insensitive](https://github.com/mehcode/config-rs/blame/4457cffb681e9765e364afb24cd28e247f9dc6e5/src/de.rs#L275) but it looks like the `release-0.13.x` branch peeled off from master a long time ago (the change was made in master 2 years ago).  Are there plans to sync with master and bring in this fix before doing another release?

matthiasbeyer:
Closes #509 

@bbaldino and @polarathene - could you please help me sort the changelog entries into the right categories, if you can spare maybe 15 minutes or so?

Don't worry if you don't have time, it will just take a bit longer then :+1: :laughing:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,195 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## Unreleased
 
+## 0.14.0 - 2024-02-01
+
+This is a maintenance release, mostly to get some dependency updates out, but
+also with some fixes and changes that have piled up over a rather long time.
+
+### Features
+
+ - [#100] Fix #99: expose file::source::FileSource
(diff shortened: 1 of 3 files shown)


matthiasbeyer:
Nah, don't worry about it ... I am faster than I thought.

PR #538, opened 2024-02-25
shuttle_secrets source

cbinzer:
Hi,

I don’t know if this is helpful. I created a new source for the [shuttle_secrets](https://docs.rs/shuttle-secrets/0.39.0/shuttle_secrets/) of the [shuttle.rs](https://www.shuttle.rs/) platform. It is a simple implementation that is based on the existing environment source. If is not useful or it should not be part of that crate then ignore the PR.

Regards
Christian


epage:
Sorry, this slipped through the cracks.

This doesn't seem to offer any additional benefit by being built-in while being fairly specialized.  I am leaning towards closing this.  If there is a reason for us to re-evaluate, please open an issue for us to discuss it there.

PR #554, opened 2024-03-27
Switch from yaml-rust to yaml-rust2 dependency

0rzech:
This is because yaml-rust is unmaintained and thus has been reported in [RUSTSEC-2024-0320](https://rustsec.org/advisories/RUSTSEC-2024-0320).

Resolves #473 .
Resolves #553 .

PR #555, opened 2024-03-27
fix: replace yaml-rust with yaml-rust2

indirection42:
#554 fix this


indirection42:
It seems it has a previour PR, so I will close this.

PR #515, opened 2024-02-01
refactor: Drop lazy_static dependency

epage:
Fixes #514

PR #571, opened 2024-07-11
Add a global_module example

robertek:
This example shows how to create global configuration module that may be used anywhere in the code with predefined interface.


softdevca:
This should be updated to use OnceLock instead of lazy_static.  See pr #515

robertek:
Thanks for pointing that out. 

I have updated that and added a comments around.

softdevca:
I made some small improvements to make clippy happy at https://github.com/fork-in-the-code/fitc-config/commit/374840a4786ed6a66eb52f5d0b321c72cb37aae9

robertek:
Thanks for the review. I've applied that change.

PR #572, opened 2024-07-22
Custom path parser that doesn't use `nom`

### LINKED ISSUE #516: Nom is included in public API
See [`ConfigError::ParsePath`](https://docs.rs/config/latest/config/enum.ConfigError.html#variant.PathParse).

This makes upgrading nom or moving away from it a breaking change.

softdevca:
The path parser was re-written to not use `nom`.  This drops a dependency, benchmarks twice as fast and fixes #516.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -180,7 +180,6 @@ dependencies = [
  "json5",
  "lazy_static",
  "log",
- "nom",
  "notify",
  "pathdiff",
  "reqwest",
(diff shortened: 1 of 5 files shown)


polarathene:
You're aware of the [current maintenance status](https://github.com/mehcode/config-rs/issues/549) of this project right?

I'm not sure if anyone with the ability to review and merge will be able to assist. When my major commitments elsewhere are resolved, there's the possibility I'd have time to review and merge some of the PRs here, but I don't think I was given the ability to publish releases.

While I have been trusted to review and merge, I'm not an official maintainer here, nor is it a responsibility I can commit to, I was granted it to help with bus-factor in case of the current situation I suppose.

---

With my present backlog, I don't see myself having time in the near future for `config-rs` (_my comment from 5 months ago in your linked issue ref is an example of how stretched I am, that I cannot make any realistic estimates for availability_).

I may be able to make an exception for a PR with high demand, but again even if I merge it.. I don't think I have the capability to make an actual release. Thus it'd be the same as using your branch as a git dependency.

---

Not to discourage the contribution as I'm sure it is very welcome. Just a little concerned with contributors putting time and effort into work that may not go anywhere if the project does not get new maintainers.

softdevca:
I am aware of the current status.  This PR is code from a fork that I was considering.  I submitted this PR in case this project is revived.   Thank you for making sure I was aware.

It feels like it's time for someone to fork.

Issue #531, opened 2024-02-03
bug: config `0.14.0` only supports lowercase fields (regression since `0.13.4`)


vladgon:
Failed to deserialize camelcase fields for yaml file
https://github.com/vladgon/rust/blob/49854f6c3d02bf7ae4a8d123ac9ecfbbe3959a56/wspace/wg_util/src/common/config/model.rs#L24
```rust
#[derive(Deserialize, Serialize, Debug)]
#[allow(non_snake_case)]
pub struct Kafka {
    pub broker: String,
    pub topic: String,
    pub pollSleep: u64,  //<---
}
```

https://github.com/vladgon/rust/blob/49854f6c3d02bf7ae4a8d123ac9ecfbbe3959a56/wspace/wg_sample_app/resources/app_config.yaml#L9
```yaml
kafka:
  broker: localhost:29092
  topic: rust
  pollSleep: 1000
```

https://github.com/vladgon/rust/blob/49854f6c3d02bf7ae4a8d123ac9ecfbbe3959a56/wspace/wg_util/src/common/config/app_config.rs#L73-L78

polarathene:
Probably have to compare the crate version differences of dependencies? Unless it was an actual internal change introduced to `config-rs` (_I recall some work was contributed to rework deserializing_).

Is this specific to YAML only? Could you try it with a different config file type?

---

If it's specific to YAML, there is plans to change the crate used: https://github.com/mehcode/config-rs/pull/474

I'm hoping to have time next week to tackle that and related PRs.

vladgon:
Tested with toml file same issue, works with 0.13.4, fails with 0.14.0

```console
Setting Tracing filter config wg_util=debug,server=debug
2024-02-03T17:19:42.381705Z DEBUG ThreadId(01) wg_util::common::config::rust_app: 29: Using config files: "/Users/vova/workspace/rust/wspace/wg_sample_app/resources/app_config.toml"    
thread 'main' panicked at wg_util/src/common/config/app_config.rs:77:82:
missing field `pollSleep`
stack backtrace:
```

polarathene:
> works with 0.13.4, fails with 0.14.0

Ok, awesome!

So now we know that it's not format specific, it's either due to a dependency update or something that changed in the actual `config-rs` code. That's a tricky one someone will need to invest time into to track down.

[...]

PR #608, opened 2024-12-05
fix: convert the keys in the array to lowercase

rise0chen:
fix https://github.com/rust-cli/config-rs/issues/609

[turn keys to lowercase to enable cross overrides](https://github.com/rust-cli/config-rs/commit/d2ee90ea23281b057dffa6515b274e5c7b82cf87) not convert the keys in the array to lowercase. This PR fix it. 

Before fix:
image

After fix:
image

### DIFF
--- a/src/path/mod.rs
+++ b/src/path/mod.rs
@@ -204,6 +204,13 @@ impl Expression {
                             Self::Identifier(key.to_lowercase()).set(target, val.clone());
                         }
                     }
+                    ValueKind::Array(ref incoming_arr) => {
+                        // Continue the deep merge
+                        for (i, val) in incoming_arr.iter().enumerate() {
+                            Self::Subscript(Box::new(self.clone()), i as isize)
+                                .set(root, val.clone());
+                        }
+                    }
 
                     _ => {
                         if let ValueKind::Table(ref mut map) = root.kind {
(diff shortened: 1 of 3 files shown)


epage:
Is there an issue that this should be linking back to?  We recommend any non-trivial (like beyond a typo) PRs to first start as issues so we make sure we have a mutual understanding the problem and the right direction to go in solving it.

When posting PRs, we also recommend the PR having at least two commits
1. A test showing the current behavior
2. The fix with the update to the test so it pass

The diff for the test makes it easy to verify tests are testing the right thing and make it very clear what the intended change is.

rise0chen:
Before fix:
image

image

rise0chen:
I have split it into two commits.

I haven't submit an issue, should I resubmit an issue?

epage:
> I haven't submit an issue, should I resubmit an issue?

Yes.  PRs are not great places for discussing problems and potential solutions.  They are much more likely to come and go and a lot of that background gets lost or bifurcated.

rise0chen:
fix https://github.com/rust-cli/config-rs/issues/609

epage:
Going to close for now as we should consider other related issues like #531 in fixing issues introduced by #354.

Issue #625, opened 2024-12-23
No option to exclude keys from case conversion


moldhouse:
Currently, it is not possible to specify a list of keys that should be excluded from case conversion.

In my case, this makes it hard to build a config from an `Environment` and a `File`, where some identifiers in the file are `kebab-case`, while other parameters are `snake_case`. 

For example, I want a `kebab-case` key and a `snake_case` configuration:

```toml
# TOML configuration
[foo.kebab-case-key]
snake_case_configuration = "bar"
```

```sh
# Environment configuration
FOO__KEBAB_CASE_KEY__SNAKE_CASE_CONFIGURATION=bar
```

While there is a method [convert_case](https://docs.rs/config/latest/config/struct.Environment.html#method.with_convert_case), it is applied to all collected keys. For this example, a possible solution would be something like:

```rust
Environment::default().convert_case_exclude_keys(Case::Kebab, ["snake_case_configuration"]);
```

epage:
I'm sorry for the delay; I'm still working on catching up from the holidays

To make this easier to process and avoid confusing, could you write this up like a bug report with a reproduction case, showing the actual behavior and reporting what the expected behavior is?

PR #626, opened 2024-12-23
feat: add case conversion strategies to include/exclude keys

moldhouse:
This PR introduces two new strategies for case conversion, which allow users to either specify a list of keys that should be converted or exclude a list of keys from conversion. This solves issue #625 that it is currently not possible to exclude some keys from case conversion.


Scutua (approved):

epage:
I'm marking this as a Draft because we don't look at PRs until a problem is understood and and a solution is agreed upon, per our [contributing guide](https://github.com/rust-cli/config-rs/blob/main/CONTRIBUTING.md#pull-requests)

PR #627, opened 2024-12-27
treat nil parents as empty tables if required

### LINKED ISSUE #624: Reading HashMap from empty json fails with type error.
version: 0.15.4

Config fails to initialize (an empty) HashMap from an empty object.

Repro: 
```
#[derive(Clone, Debug, Serialize, Deserialize)]
#[serde(rename_all = "camelCase")]
pub struct Config {
    pub data: HashMap<String, String>,
}
...
 let cfg = config::Config::builder()
            .add_source(config::File::from_str(
                //r#"{"data": {"a": "b"}}"#,  //  it works fine
                r#"{"data": {}}"#,                // it fails
                config::FileFormat::Json,
            ))
            .build()
            .unwrap();
        let cfg = cfg.try_deserialize::<Config>();
        log::info!("configuration: {:#?}", cfg);
```

Expected behavior: data is initialized to an empty HashMap.
Actual result: `Err( invalid type: unit value, expected a map )`

sunshowers:
Encountered this regression via nextest's test suite (https://github.com/nextest-rs/nextest/pull/2001). If a table is empty, it would be treated as a unit value. If attempted to be deserialized via a `Default` impl, this would lead to deserialization failing with an error like

```console
profile.default-miri: invalid type: unit value, expected struct CustomProfileImpl
```

A bisect appears to indicate that ec36bff45eb339077e23d47d2edabc0bcd03cf6c is responsible.

I've attempted to restore the old behavior of putting in an empty table, specifically this section:

https://github.com/rust-cli/config-rs/commit/ec36bff45eb339077e23d47d2edabc0bcd03cf6c#diff-c5423e2d2d6c87501239c0304c0f496742e00440defdd20368cf548ba42ab184L175-L178

I'm happy to make changes if there's a better approach.

I've also added some tests for a few situations around empty tables. In this case only the last one (`profile.baz`) regressed, but I've also added tests for a couple other cases.

Fixes #624

### DIFF
--- a/src/path/mod.rs
+++ b/src/path/mod.rs
@@ -150,12 +150,16 @@ impl Expression {
         let parent = self.get_mut_forcibly(root);
         match value.kind {
             ValueKind::Table(ref incoming_map) => {
+                // If the parent is nil, treat it as an empty table
+                if matches!(parent.kind, ValueKind::Nil) {
+                    *parent = Map::<String, Value>::new().into();
+                }
+
                 // Continue the deep merge
                 for (key, val) in incoming_map {
                     Self::root(key.clone()).set(parent, val.clone());
                 }
             }
-
             _ => {
                 *parent = value;
             }
--- a/src/path/mod.rs
+++ b/src/path/mod.rs
@@ -150,8 +150,9 @@ impl Expression {
         let parent = self.get_mut_forcibly(root);
         match value.kind {
             ValueKind::Table(ref incoming_map) => {
-                // If the parent is nil, treat it as an empty table
-                if matches!(parent.kind, ValueKind::Nil) {
+                // If the parent is not a table, overwrite it, treating it as a
+                // table
+                if !matches!(parent.kind, ValueKind::Table(_)) {
                     *parent = Map::<String, Value>::new().into();
                 }
 
--- a/tests/testsuite/merge.rs
+++ b/tests/testsuite/merge.rs
@@ -1,3 +1,5 @@
+use snapbox::{assert_data_eq, prelude::*, str};
+
 use config::{Config, File, FileFormat, Map};
 
 #[test]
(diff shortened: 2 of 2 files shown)


Holzhaus:
Great, I can confirm that this fixes the regression.

sunshowers:
@epage apologies about pinging directly -- any chance you can look at this? this is a pretty rough regression, I'm currently planning to pin nextest to an earlier version of config-rs to avoid it

Holzhaus:
I also had to reject the depency update for one of my projects due to this regression: https://github.com/Holzhaus/helicon/actions/runs/12540222602/job/34967265515?pr=49

epage:
Please keep in mind we just had holidays in the US and I'm still slowly catching up while handling other responsibilities.

As for this, two quick thoughts
- this doesn't seem toml related, so shouldn't be in the toml tests. Outside of format tests, json is our go-to
- It is a big help to split fixes into two commits, the first that just adds a passing test, showing the existing behavior, the second fixing the bug with a test update.

In this model, the diff of the test between the commits shows how behavior changed, showing the test verifies the right behavior and clearly communicates to reviewers and any one else how the behavior changed

sunshowers:
Thanks -- addressing them shortly.

sunshowers:
All right, let me know if there are any other changes. Thanks!

(Totally understand the delay btw! Just wanted to ensure the severity was understood)

src/path/mod.rs
```diff
@@ -150,12 +150,16 @@ impl Expression {
         let parent = self.get_mut_forcibly(root);
         match value.kind {
             ValueKind::Table(ref incoming_map) => {
+                // If the parent is nil, treat it as an empty table
+                if matches!(parent.kind, ValueKind::Nil) {
+                    *parent = Map::<String, Value>::new().into();
+                }
+
```
epage, line 157: If I'm reading the old code correctly, it overwrites `parent` if its not a table.  is there a reason this is limited to `Nil`?

sunshowers: Hmmm, good point. I'm not quite sure what behavior's expected there, let me try a few things.

sunshowers: OK, I've spent the last hour or so looking at this code, and I think that:

* the previous code pre-regression and the code after this change do roughly the same thing
* they both might be slightly wrong?

So backing up a little, the problem we're hitting here is specifically around the fact that:

* if a table is empty
* and if we expect that the `Default` impl is used

then before https://github.com/rust-cli/config-rs/commit/ec36bff45eb339077e23d47d2edabc0bcd03cf6c, it's resolved as an empty table, but after that change, it's resolved to be nil.

It's worth looking at this case-by-case, across three worlds:

1. Before the regression (i.e. v0.15.2)
2. After the regression in https://github.com/rust-cli/config-rs/commit/ec36bff45eb339077e23d47d2edabc0bcd03cf6c
3. After my PR

## Before the regression

In [lines 175-178 on the before side](https://github.com/rust-cli/config-rs/commit/ec36bff45eb339077e23d47d2edabc0bcd03cf6c) (the "Pull out another table" block), on attempting to set a table, the behavior was:

* **if previously not present:** make an empty table
* **if previously present but not a table:** retain as-is
* **if previously a table:** retain as-is

Then, in lines 181, we continue the deep merge, and the recursive function would cause any non-table values to be overwritten via the recursive call's "Ensure that root is a table" block.

However, note that line 181 is only active if `incoming_map` is non-empty. If it's empty, then no overwrite happens and the previous value is retained. This means that this breaks down into 2 worlds.

* If the incoming table is non-empty, the value is always overwritten with a table.
* If the incoming table is empty, either an empty table is made or the old value is retained as-is (as per the list above)

## After the regression

After https://github.com/rust-cli/config-rs/commit/ec36bff45eb339077e23d47d2edabc0bcd03cf6c, the `get_mut_forcibly` will cause `Nil` to be set via https://github.com/rust-cli/config-rs/blob/9a0c1a16aa4b092f1ff3220243fc38465fd10271/src/path/mod.rs#L101-L103. So, after `get_mut_forcibly`:

* **if previously not present:** make a *nil value*
* **if previously present but not a table:** retain as-is
* **if previously a table:** retain as-is

Then, lines 168-171 continue the deep merge -- but again, only if the incoming map has at least one key.

* If it does, `get_mut_forcibly` will force it to be a table via https://github.com/rust-cli/config-rs/blob/9a0c1a16aa4b092f1ff3220243fc38465fd10271/src/path/mod.rs#L95-L97.
* If not, then the conditions above apply.

## After this PR

As it currently stands, my PR sets a table if and only if the value was nil (which *generally* indicates that `get_mut_forcibly` just set that nil, but not *always*). This means that the effective behavior of phase 1 is:

* **if previously not present:** make an empty table
* **if previously present and nil:** make an empty table
* **if previously present but not a table and not nil:** retain as-is
* **if previously a table:** retain as-is

So there is a behavior difference here, which is that if the value changed from exactly `Nil` to an empty table, we now set an empty table. And indeed, if you make [this change](https://gist.github.com/sunshowers/97fd5ebc300294115ee16031b1ac567e#file-change-diff-L7) to my test, it fails with 0.15.2 but passes after my PR.

## But it's weird to treat nil as special

At least from this perspective, it seems to me that there isn't anything special about nil/null. Changing an i64 or an array into an empty table should also result in an empty table, I think it's fair to say.

Really, I think that there's no reason to treat empty tables any different from non-empty ones.

So yes, I think we should do what you said. But that would be a notable behavior change compared to v0.15.2.

---

Here's a table which summarizes all this.

* **existing** is the value already present
* **incoming** is the value being merged
* **v0.15.2** is the behavior before the regression
* **main** is the behavior after the regression, and on current main
* **PR** is the behavior in this PR
* **proposed** is what I think you're proposing, and I agree, but is a behavior change

|   existing   |     incoming    |     v0.15.2    |      main      |       PR       |    proposed    |
|:------------:|:---------------:|:--------------:|:--------------:|:--------------:|:--------------:|
| non-existent |   empty table   |   empty table  |       nil      |   empty table  |   empty table  |
| non-existent | non-empty table | incoming table | incoming table | incoming table | incoming table |
|     table    |   empty table   |   table as-is  |   table as-is  |   table as-is  |   table as-is  |
|     table    | non-empty table |   deep merge   |   deep merge   |   deep merge   |   deep merge   |
|      nil     |   empty table   |       nil      |       nil      |   empty table  |   empty table  |
|      nil     | non-empty table | incoming table | incoming table | incoming table | incoming table |
|  other value |   empty table   |   other value  |   other value  |   other value  |   empty table  |
|  other value | non-empty table | incoming table | incoming table | incoming table | incoming table |

epage: Do you have test cases for each of these situations that could be added to this PR?

Trying to play around with ideas to even figure out what I think of any of this.  I suspect we have some wiggle room in some cases but it can be hard to predict what will break people.

sunshowers: @epage sorry for the delay, other things came up -- have updated the PR with the proposed change, as well as a test for all of these cases.

spencewenski:
Thanks @sunshowers and @epage for getting this fixed!

PR #630, opened 2025-01-03
fix(file): Combine path and pathbuf

allenanswerzq:
Small fix to combine the path and pathbuf into one implementaion.


epage:
This is described as a fix which would imply end-user effect but the motivation seems to be focused on a refactor.

I tend to avoid blanket implementations because of how limiting they are and would prefer to stick to individual implementations unless there is a motivating reason.

allenanswerzq:
yeah, its make sense, will close this one now.

PR #632, opened 2025-01-09
Add `try_deserialize` deserialization path error tracking

allevo:
Hi!
I'm using `config-rs`, but I'm having some difficulties finding where my JSON file lacks fields.
I have a nested configuration struct like the one below:
```rust
#[derive(Debug, Deserialize)]
struct InnerSettings {
    #[allow(dead_code)]
    value: u32,
    #[allow(dead_code)]
    value2: u32,
}

#[derive(Debug, Deserialize)]
struct Settings {
    #[allow(dead_code)]
    inner: InnerSettings,
}
```

NB: my use case involves more complex structures

So, if you try to deserialize a wrong JSON, the error will contains the JSON path:
```rust

let c = Config::builder()
    .add_source(File::from_str(
        r#"
{
"inner": { "value": 42 }
}
    "#,
        FileFormat::Json,
    ))
    .build()
    .unwrap();

let without_path = c.clone().try_deserialize::<Settings>();
assert_data_eq!(
    without_path.unwrap_err().to_string(),
    str!["inner: missing field `value2`"] // prepended with JSON path
);
```

This PR:
- adds [`serde_path_to_error`](https://crates.io/crates/serde_path_to_error)
- add a JSON path if not empty in front of the error.
- adds a test
- add method documentation

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -135,6 +135,7 @@ indexmap = { version = "2.2", features = ["serde"], optional = true }
 convert_case = { version = "0.6", optional = true }
 pathdiff = "0.2"
 winnow = "0.6.20"
+serde_path_to_error = "0.1"
 
 [dev-dependencies]
 serde_derive = "1.0"
--- a/src/config.rs
+++ b/src/config.rs
@@ -2,6 +2,7 @@ use std::fmt::Debug;
 
 use serde::de::Deserialize;
 use serde::ser::Serialize;
+use serde_path_to_error::{Deserializer, Track};
 
 use crate::builder::{ConfigBuilder, DefaultState};
 use crate::error::{ConfigError, Result};
@@ -142,7 +143,20 @@ impl Config {
 
     /// Attempt to deserialize the entire configuration into the requested type.
     pub fn try_deserialize<'de, T: Deserialize<'de>>(self) -> Result<T> {
-        T::deserialize(self)
+        let mut track = Track::new();
+        match T::deserialize(Deserializer::new(self, &mut track)) {
+            Ok(t) => Ok(t),
+            Err(ConfigError::Message(e)) => {
+                let path = track.path();
+                if path.iter().count() == 0 {
+                    return Err(ConfigError::Message(e));
+                }
+                Err(ConfigError::Message(format!(
+                    "failed reading field `{path}`: {e}"
+                )))
+            }
+            Err(err) => Err(err),
+        }
     }
 
     /// Attempt to serialize the entire configuration from the given type.
--- a/tests/testsuite/errors.rs
+++ b/tests/testsuite/errors.rs
@@ -316,3 +316,39 @@ fn test_error_root_not_table() {
         },
     }
 }
+
+#[test]
+#[cfg(feature = "json")]
+fn test_json_error_with_path() {
+    #[derive(Debug, Deserialize)]
+    struct InnerSettings {
+        #[allow(dead_code)]
+        value: u32,
+        #[allow(dead_code)]
+        value2: u32,
+    }
+
+    #[derive(Debug, Deserialize)]
+    struct Settings {
+        #[allow(dead_code)]
+        inner: InnerSettings,
+    }
+
+    let c = Config::builder()
+        .add_source(File::from_str(
+            r#"
+{
+    "inner": { "value": 42 }
+}
+        "#,
+            FileFormat::Json,
+        ))
+        .build()
+        .unwrap();
+
+    let with_path = c.clone().try_deserialize::<Settings>();
+    assert_data_eq!(
+        with_path.unwrap_err().to_string(),
+        str!["missing field `value2`"]
+    );
+}
--- a/tests/testsuite/errors.rs
+++ b/tests/testsuite/errors.rs
@@ -341,7 +341,7 @@ fn test_json_error_with_path() {
     "inner": { "value": 42 }
 }
         "#,
-            FileFormat::Json,
+            FileFormat::Json5,
         ))
         .build()
         .unwrap();
@@ -349,6 +349,6 @@ fn test_json_error_with_path() {
     let with_path = c.clone().try_deserialize::<Settings>();
     assert_data_eq!(
         with_path.unwrap_err().to_string(),
-        str!["missing field `value2`"]
+        str!["failed reading field `inner`: missing field `value2`"]
     );
 }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -254,6 +254,7 @@ dependencies = [
  "serde",
  "serde_derive",
  "serde_json",
+ "serde_path_to_error",
  "snapbox",
  "temp-env",
  "tokio",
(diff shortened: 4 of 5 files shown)


tests/testsuite/file_json_error_path.rs
```diff
@@ -0,0 +1,46 @@
+#![cfg(all(feature = "json5", feature = "path_to_error"))]
+
+use serde_derive::Deserialize;
```
epage, line 3: Please put this in the `errors.rs` file

Cargo.toml
```diff
@@ -120,6 +120,7 @@ convert-case = ["convert_case"]
 preserve_order = ["indexmap", "toml?/preserve_order", "serde_json?/preserve_order", "ron?/indexmap"]
 async = ["async-trait"]
 toml = ["dep:toml"]
+path_to_error = ["serde_path_to_error"]
```
epage, line 123: I would be interested in seeing a proposal for what it would look like to have this built-in

tests/testsuite/file_json_error_path.rs
```diff
@@ -0,0 +1,46 @@
...
+        Ok(_) => panic!("expected a type error"),
+        Err(err) => {
+            let path = err.path().to_string();
+            assert_eq!(path, "inner.value");
+
+            let s = format!("{err}");
+            assert_eq!(
+                s,
+                "inner.value: invalid type: unit value, expected a string for key `inner.value`"
+            );
```
epage, line 43: iirc the other tests use `snapbox` for assertions

tests/testsuite/file_json_error_path.rs
allevo: addressed!

Cargo.toml
allevo: Do you mean without any external dependency? I think it is doable, but probably we will copy & paste from the other repository.
Or maybe I miss-interpreted the point. Can you elaborate more?
thanks

tests/testsuite/file_json_error_path.rs
```diff
@@ -0,0 +1,46 @@
...
+        }
```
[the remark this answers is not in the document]
allevo: Ops. Sorry, my bad.
Anyway, how do you suggest to proceed?
I can create a first commit a test that uses `try_deserialize` where the assertion fails (it miss the path). I can put a `#[ignore]` as well.
After, I can commit the implementation, changing also the test to point to `try_deserialize_with_error_path`.

Let me know, and sorry again.

tests/testsuite/file_json_error_path.rs
allevo: done.

Cargo.toml
epage: Without it being optional and baked into the existing error reporting

tests/testsuite/file_json_error_path.rs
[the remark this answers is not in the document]
epage: Please see https://github.com/rust-cli/config-rs/blob/main/CONTRIBUTING.md#pull-requests

We include https://github.com/clap-rs/clap/pull/5520 as an example (I need to find a better one ...)

The test in the first commit should pass and should not be using `#[ignore]`.  The intent behind this is to show how the behavior changed with the feature work and `#[ignore]` masks that.

epage:
And feel free to squash commits that are just for review feedback

Cargo.toml
allevo: remove optional

tests/testsuite/errors.rs
```diff
@@ -349,6 +349,6 @@ fn test_json_error_with_path() {
     let without_path = c.clone().try_deserialize::<Settings>();
     assert_data_eq!(
         without_path.unwrap_err().to_string(),
-        str!["missing field `value2`"]
+        str!["inner: missing field `value2`"]
```
epage, line 352: Now the question is how to format the message.

I feel like this doesn't make clear what `inner` is.

One idea: "failed reading field `inner`: missing field `value2`"

tests/testsuite/file_json_error_path.rs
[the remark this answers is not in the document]
allevo: I think I write the commit in the right way. Let me know.

tests/testsuite/errors.rs
allevo: make sense

allevo: done

epage: In looking at `fn get`, I saw that we have some level of path tracking.  In those cases, we append the path with `for path foo.bar`.  Should we be consistent one way or the other?

src/config.rs
```diff
@@ -142,7 +143,20 @@ impl Config {
 
     /// Attempt to deserialize the entire configuration into the requested type.
     pub fn try_deserialize<'de, T: Deserialize<'de>>(self) -> Result<T> {
-        T::deserialize(self)
+        let mut track = Track::new();
+        match T::deserialize(Deserializer::new(self, &mut track)) {
```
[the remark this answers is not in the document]
epage: (sorry, just now noticed that this was a whole-document case separate from the path look up case)

tests/testsuite/errors.rs
epage: I suspect they don't overlap.

With `get`, we extract a `Value` and then deserialize that.  Any path `serde_path_to_error` provides will only be within that `Value`.  The `path` the user passes in will be separate and instead we'll want to merge the two.

allevo:
```
missing field `value2` at path `inner`
``` 

?

tests/testsuite/errors.rs
```diff
@@ -316,3 +316,39 @@ fn test_error_root_not_table() {
         },
     }
 }
+
+#[test]
+#[cfg(feature = "json5")]
```
epage, line 321: I recently made all of the tests consistently use `json` unless they are format specific.  Is there a reason this uses json5?

tests/testsuite/errors.rs
epage: It might be good to double check additional error cases to make sure we aren't being biased by what wording works for the current one

tests/testsuite/errors.rs
allevo: done

tests/testsuite/errors.rs
allevo: I tested locally other cases:
```rust

#[derive(Debug, Deserialize)]
struct InnerSettings {
    #[allow(dead_code)]
    value: u32,
    #[allow(dead_code)]
    value2: u32,
}

#[derive(Debug, Deserialize)]
struct Settings {
    #[allow(dead_code)]
    inner: InnerSettings,
}

let c = Config::builder()
    .add_source(File::from_str(
        r#"
{
"inner": { "value": 42 }
}
    "#,
        FileFormat::Json5,
    ))
    .build()
    .unwrap();

let with_path = c.clone().try_deserialize::<Settings>();
assert_data_eq!(
    with_path.unwrap_err().to_string(),
    str!["missing field `value2` in path `inner`"]
);

let c = Config::builder()
    .add_source(File::from_str(
        r#"
{
"inner": { "value": "abc" }
}
    "#,
        FileFormat::Json5,
    ))
    .build()
    .unwrap();

let with_path = c.clone().try_deserialize::<Settings>();
assert_data_eq!(
    with_path.unwrap_err().to_string(),
    str!["invalid type: string \"abc\", expected an integer for key `inner.value`"]
);
```

So:
- if the key is missing: "missing field `value2` in path `inner`"
- if the value isn't "castable": "invalid type: string \"abc\", expected an integer for key `inner.value`"

it seems to be consistent.

Let me know.

tests/testsuite/errors.rs
allevo: done

tests/testsuite/errors.rs
```diff
@@ -341,14 +341,14 @@ fn test_json_error_with_path() {
     "inner": { "value": 42 }
 }
         "#,
-            FileFormat::Json,
+            FileFormat::Json5,
```
epage, line 344: This commit changes it back to json5.  Merge conflict problem?

tests/testsuite/errors.rs
epage: That we already have the path for the other error is suspicious and implies we already have path tracking we can take advantage of.

allevo: This is why I wrote this comment https://github.com/rust-cli/config-rs/pull/632#discussion_r1909254010.
Currently, the missing field is not supported because the implementation only track the cast issues and not the deserialization error (which is hard to implement without external library).

epage: Sorry, I misunderstood what you were referring to.  We have separate path tracking in `get` on top of whatever else happens and I thought you were referring to that.

I have a branch with the path tracking without an external library.  I'm trying to add `origin` tracking as well but might give up for now.

epage: See #635

allevo: .... ok ....
Thanks anyway for the feedback.

Issue #579, opened 2024-10-21
Panics when any environment variable contains non-unicode characters


printfn:
This crate causes a panic if any environment variable contains non-unicode characters, such as when `PWD` points at a non-unicode directory, e.g. when running `mkdir $'\xff'`.

https://github.com/mehcode/config-rs/blob/a38387955622e6b8d48a49352629cd329ad6dc29/src/env.rs#L239

This line needs to be changed to `env::vars_os()`, with appropriate Unicode escaping and/or filtering of unencodable environment variables.

Downstream issue: https://github.com/atuinsh/atuin/issues/2428

Jongy:
I just hit this issue myself while working with atuin in a non-UTF8 directory 😅 just like your case. The fix is trivial but I'm unsure if it's correct in all cases. I'll push a PR with it later.

epage:
Would appreciate it, thanks!

PR #640, opened 2025-02-05
Handle non Unicode env keys/values

Jongy:
(disregard the code / commit message - I'll clean up & add tests once we agree on the implementation :smile:  )

I hit the same issue as described in the linked ticket - with key `PWD` having a non-UTF-8 value ("invalid" from now on).

Technically, keys can be invalid too in Linux. The naive implementation I provided here filters out all invalid keys & values. Keys in config-rs are `String`s so it always makes sense to filter invalid ones, as they won't be used by this library AFAIU.
As for values - I'm unsure if plainly ignoring invalid values is okay. If the value is set in a key that would have been used by the current config consumer, then ignoring it (as if the user did not provide a value through env) is not okay.
Ideally - the logic would be:
1. Valid key & value - continue
2. Invalid key - drop
3. Invalid value - drop if key is not consumed, otherwise panic.

Let me know what you think :) I can try looking into point 3.

Resolves: https://github.com/rust-cli/config-rs/issues/579

### DIFF
--- a/src/env.rs
+++ b/src/env.rs
@@ -1,4 +1,5 @@
 use std::env;
+use std::ffi::OsString;
 
 #[cfg(feature = "convert-case")]
 use convert_case::{Case, Casing};
@@ -7,6 +8,7 @@ use crate::error::Result;
 use crate::map::Map;
 use crate::source::Source;
 use crate::value::{Value, ValueKind};
+use crate::ConfigError;
 
 /// An environment source collects a dictionary of environment variables values into a hierarchical
 /// config Value type. We have to be aware how the config tree is created from the environment
@@ -243,10 +245,16 @@ impl Source for Environment {
             .as_ref()
             .map(|prefix| format!("{prefix}{prefix_separator}").to_lowercase());
 
-        let collector = |(key, value): (String, String)| {
+        let collector = |(key, value): (OsString, OsString)| {
+            let key = match key.into_string() {
+                Ok(key) => key,
+                // Key is not valid unicode, skip it
+                Err(_) => return Ok(()),
+            };
+
             // Treat empty environment variables as unset
             if self.ignore_empty && value.is_empty() {
-                return;
+                return Ok(());
             }
 
             let mut key = key.to_lowercase();
@@ -260,9 +268,17 @@ impl Source for Environment {
                     }
                 } else {
                     // Skip this key
-                    return;
+                    return Ok(());
                 }
             }
+
+            // At this point, we don't know if the key is required or not.
+            // Therefore if the value is not a valid unicode string, we error out.
+            let value = value.into_string().map_err(|os_string| {
+                ConfigError::Message(format!(
+                    "env variable {key:?} contains non-Unicode data: {os_string:?}"
+                ))
+            })?;
 
             // If separator is given replace with `.`
             if !separator.is_empty() {
@@ -308,12 +324,18 @@ impl Source for Environment {
             };
 
             m.insert(key, Value::new(Some(&uri), value));
+
+            Ok(())
         };
 
         match &self.source {
-            Some(source) => source.clone().into_iter().for_each(collector),
-            None => env::vars().for_each(collector),
-        }
+            Some(source) => source
+                .clone()
+                .into_iter()
+                .map(|(key, value)| (key.into(), value.into()))
+                .try_for_each(collector),
+            None => env::vars_os().try_for_each(collector),
+        }?;
 
         Ok(m)
     }
--- a/tests/testsuite/env.rs
+++ b/tests/testsuite/env.rs
@@ -751,7 +751,6 @@ mod unicode_tests {
     }
 
     #[test]
-    #[should_panic(expected = r#"`Result::unwrap()` on an `Err` value: "\xFF""#)]
     fn test_invalid_unicode_key_ignored() {
         temp_env::with_vars(
             vec![
@@ -767,7 +766,6 @@ mod unicode_tests {
     }
 
     #[test]
-    #[should_panic(expected = r#"`Result::unwrap()` on an `Err` value: "\xFF""#)]
     fn test_invalid_unicode_value_filtered() {
         temp_env::with_vars(
             vec![
@@ -787,12 +785,21 @@ mod unicode_tests {
     }
 
     #[test]
-    #[should_panic(expected = r#"`Result::unwrap()` on an `Err` value: "\xFF""#)]
     fn test_invalid_unicode_value_not_filtered() {
         temp_env::with_vars(
             vec![("invalid_value1", Some(make_invalid_unicode_os_string()))],
             || {
-                Environment::default().collect().unwrap();
+                let result = Environment::default().collect();
+
+                #[cfg(unix)]
+                let expected =
+                    str![[r#"env variable "invalid_value1" contains non-Unicode data: "/xFF""#]];
+                #[cfg(windows)]
+                let expected = str![[
+                    r#"env variable "invalid_value1" contains non-Unicode data: "/u{d800}""#
+                ]];
+
+                assert_data_eq!(result.unwrap_err().to_string(), expected);
             },
         );
     }
--- a/tests/testsuite/env.rs
+++ b/tests/testsuite/env.rs
@@ -721,3 +721,79 @@ fn test_parse_uint_default() {
     let config: TestUint = config.try_deserialize().unwrap();
     assert_eq!(config.int_val, 42);
 }
+
+#[cfg(any(unix, windows))]
+#[cfg(test)]
+mod unicode_tests {
+    use std::ffi::OsString;
+
+    use super::*;
+
+    fn make_invalid_unicode_os_string() -> OsString {
+        let string = {
+            #[cfg(unix)]
+            {
+                use std::os::unix::ffi::OsStringExt;
+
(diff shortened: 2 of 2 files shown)


src/env.rs
```diff
@@ -243,6 +244,21 @@ impl Source for Environment {
...
+            let key = if let Ok(key) = key.into_string() {
+                key
+            } else {
+                return None;
+            };
+            let value = if let Ok(value) = value.into_string() {
+                value
+            } else {
+                return None;
+            };
```
epage, line 257: Does the following work or does it run into type inference issues?
```suggestion
            let key = key.into_string().ok()?;
            let value = value.into_string().ok()?;
```

src/env.rs
```diff
@@ -312,7 +328,7 @@ impl Source for Environment {
 
         match &self.source {
             Some(source) => source.clone().into_iter().for_each(collector),
-            None => env::vars().for_each(collector),
+            None => env::vars_os().filter_map(filter_utf8).for_each(collector),
```
epage, line 331: > Technically, keys can be invalid too in Linux. The naive implementation I provided here filters out all invalid keys & values. Keys in config-rs are Strings so it always makes sense to filter invalid ones, as they won't be used by this library AFAIU.
> As for values - I'm unsure if plainly ignoring invalid values is okay. If the value is set in a key that would have been used by the current config consumer, then ignoring it (as if the user did not provide a value through env) is not okay.
> Ideally - the logic would be:
> 
>     Valid key & value - continue
>     Invalid key - drop
>     Invalid value - drop if key is not consumed, otherwise panic.

Panic would never be the right answer.  I could see us erroring

As for knowing when they are used, `collect` at least knows if they should be collected (have the prefix if set) but that doesn't give us the full story.  To do so, we'd need a new `ValueKind` to defer out the error as long as possible but that would be a breaking change.

Given the options, maybe
- Filter invalid keys
- Error when collecting invalid values *if* they aren't filtered out
- If someone cares about deferring errors further, open an issue for us to track that for the next breaking change

src/env.rs
Jongy: Yes, type inference works, but I think that the changing following your suggestion make it obsolete.

src/env.rs
Jongy: Ok, I implemented these steps.

src/env.rs
```diff
@@ -7,6 +10,19 @@ use crate::error::Result;
 use crate::map::Map;
 use crate::source::Source;
 use crate::value::{Value, ValueKind};
+use crate::ConfigError;
+
+/// Failed to decode an environment variable value as a Unicode string.
+#[derive(Debug)]
+pub struct InvalidUnicode(OsString);
```
Jongy, line 17: I first tried to find a stdlib error to return, but the APIs returning the conversion error of OsString -> String seem to be OS-specific, so I added this generic one. (that's also why I report it via `ConfigError::Foreign`, perhaps another variant is preferred? and I could move this to `error.rs`)

src/env.rs
```diff
@@ -308,12 +336,18 @@ impl Source for Environment {
...
         };
 
         match &self.source {
-            Some(source) => source.clone().into_iter().for_each(collector),
-            None => env::vars().for_each(collector),
-        }
+            Some(source) => source
+                .clone()
+                .into_iter()
+                .map(|(key, value)| (OsString::from(key), OsString::from(value)))
```
Jongy, line 347: I don't have a good idea on how to avoid it. On Linux, at least, this should be zero cost as the underlying `Vec<u8>` is just moved... but on Windows it might not be

tests/testsuite/env.rs
```diff
@@ -721,3 +721,62 @@ fn test_parse_uint_default() {
     let config: TestUint = config.try_deserialize().unwrap();
     assert_eq!(config.int_val, 42);
 }
+
+#[cfg(test)]
```
Jongy, line 725: all 3 tests are failing before the fixing commit :+1:

epage:
I'm sorry, I overlooked that I was reviewing two different PRs (#641) for the same problem.

The tests in #641 were able to also cover Windows which could be nice.

src/env.rs
```diff
@@ -260,10 +282,16 @@ impl Source for Environment {
...
-                    return;
+                    return Ok(());
                 }
             }
 
+            // At this point, we don't know if key is required or not. Therefore the value is not a valid unicode
+            // string we error out.
+            let value = value
+                .into_string()
+                .map_err(|os_string| ConfigError::Foreign(InvalidUnicode(os_string).into()))?;
```
epage, line 293: Do we gain much benefit from creating a one-off error just for it to be errased when we can use `ConfigError::Message` like in #641?

src/env.rs
epage: On Windows, an `OsString` is still a `Vec<u8>` and is full.  The documentation for the `From` impl even says

> This conversion does not allocate or copy memory.

tests/testsuite/env.rs
epage: Every commit should be atomic which includes passing tests.

In this case, the change is simple, just adding a `#[should_panic]` to each test (preferably with the message set so we know it isn't an assert).  We specifically encourage this splitting of commits in our [contributor guide](https://github.com/rust-cli/config-rs/blob/main/CONTRIBUTING.md#process).

tests/testsuite/env.rs
```diff
@@ -772,7 +773,9 @@ mod unix_test {
             || {
                 let result = Environment::default().collect();
 
-                assert!(matches!(result, Err(ConfigError::Foreign(_))));
+                assert!(
+                    matches!(result, Err(ConfigError::Foreign(e)) if e.downcast_ref::<InvalidUnicode>().is_some())
```
epage, line 777: We should have snapshot tests of the rendered error so we can validate how this fully looks to the user

src/env.rs
Jongy: It allows the user to create specific handling for the error type. (I'll add the faulting key as well; so for example - remove the key from env and retry)

epage: While they can, will they? We can always migrate from few guarentees to more later but the other way is a breaking change. When we speculatively add features, it is also hard to know if people rely on them and why to then plan breaking changes.

src/env.rs
Jongy: Cool. I thought it'd be u16 to match Windows UTF-16, but you're right :) Resolving this, then.

tests/testsuite/env.rs
Jongy: Cool - I read that part in the contributor guide but didn't think of `should_panic`. Practically like xfailing the test. Makes sense, added it.

tests/testsuite/env.rs
Jongy: Done

Jongy (commented):
Thanks for the review @epage , I addressed all comments.

> The tests in https://github.com/rust-cli/config-rs/pull/641 were able to also cover Windows which could be nice.

Fixed the tests here to support Windows. I haven't set up Wine to run them so I'll wait to see how it goes in the CI.

tests/testsuite/env.rs
```diff
@@ -790,12 +789,15 @@ mod unicode_tests {
...
     fn test_invalid_unicode_value_not_filtered() {
         temp_env::with_vars(
             vec![("invalid_value1", Some(make_invalid_unicode_os_string()))],
             || {
-                Environment::default().collect().unwrap();
+                let result = Environment::default().collect();
+
+                assert!(
+                    matches!(result, Err(ConfigError::Message(msg)) if msg == r#"key "invalid_value1" is not valid Unicode: "\xFF""#)
+                );
```
epage, line 800: Could you instead assert this using a snapshot test, like https://github.com/rust-cli/config-rs/blob/370c939a1ad6d5b52dcfaa413604ba921773515c/tests/testsuite/errors.rs#L46-L49

Jongy: Done. Also fixed the windows test, I think.

src/env.rs
```diff
@@ -260,10 +268,16 @@ impl Source for Environment {
...
                     // Skip this key
-                    return;
+                    return Ok(());
                 }
             }
 
+            // At this point, we don't know if the key is required or not.
+            // Therefore if the value is not a valid unicode string, we error out.
+            let value = value.into_string().map_err(|os_string| {
+                ConfigError::Message(format!("key {key:?} is not valid Unicode: {os_string:?}"))
```
epage, line 278: Since we already are making a change, we should fix this error message.  THis implies the key is invalid and not the value it points to.  We also don't give the context for this being an env variable.

maybe something like "env variable `{key}` contains non-UTF 8 data: {osstring:?}"

Jongy: Np. Shouldn't it remain `Unicode` and not `UTF-8`? I deliberately changed all references to `Unicode` because [OsStr.to_str](https://doc.rust-lang.org/std/ffi/struct.OsStr.html#method.to_str) uses this term.

epage: Sure

Jongy: I also went with `{key:?}` so it's properly quoted.

epage:
Thanks!

Jongy:
Thanks again for the review, and thanks for releasing, I'll proceed to update atuin to use the new version.

PR #641, opened 2025-02-05
fix: Handle non-UTF-8 environment variables

printfn:
This fixes non-UTF-8 environment variables. It filters out any variables with non-UTF-8 keys, and returns an error if attempting to consume a value that's not UTF-8 compatible. I've also added some unit tests for this.

Resolves: https://github.com/rust-cli/config-rs/issues/579

### DIFF
--- a/tests/testsuite/env.rs
+++ b/tests/testsuite/env.rs
@@ -1,3 +1,5 @@
+use std::ffi;
+
 use serde_derive::Deserialize;
 use snapbox::{assert_data_eq, str};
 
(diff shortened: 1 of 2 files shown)


printfn:
There we go, Windows tests should be fixed now.

tests/testsuite/env.rs
```diff
@@ -721,3 +723,49 @@ fn test_parse_uint_default() {
     let config: TestUint = config.try_deserialize().unwrap();
     assert_eq!(config.int_val, 42);
 }
+
+#[test]
+fn test_non_utf8() {
+    // make a non-utf8 OsString
+    fn make_os_string(prefix: &str) -> ffi::OsString {
```
epage, line 730: Can you make this more explicit, like `append_non_utf8`?

tests/testsuite/env.rs
```diff
@@ -721,3 +723,49 @@ fn test_parse_uint_default() {
     let config: TestUint = config.try_deserialize().unwrap();
     assert_eq!(config.int_val, 42);
 }
+
+#[test]
+fn test_non_utf8() {
```
epage, line 728: Can we split this into separate key/value tests?

epage, line 728: Would you be willing to split this into multiple commits
1. Add the test with a `should_panic` so it passes
2. Change behavior and remove `should_panic` so it passes

In cargo, we've found structuring PRs like this to be a big help
- tests the tests (was that actually invalid utf8 data?)
- Makes clear the intent of the change to reviewers and the community (less important here as its more obvious)

epage:
I'm sorry, I overlooked that I was reviewing two different PRs (#640) for the same problem.

Did you close this because of the other one?

Issue #547, opened 2024-03-15
Unreachable code fault at config-0.14.0/src/file/format/yaml.rs:53:26


mjaakkol:
I'm getting above error as
```
thread 'XXX' panicked at /home/username/.cargo/registry/src/index.crates.io-6f17d22bba15001f/config-0.14.0/src/file/format/yaml.rs:53:26:
internal error: entered unreachable code
```

Yaml file looks like
```
link:
  null:
    name: "null"
  tcp:
    address: "localhost:1234"
```

My structure definitions for the section are:
```
#[derive(serde::Deserialize, Clone)]
pub struct LinkSettings {
    pub serial: Option<SerialSettings>,
    pub unix: Option<UnixSettings>,
    pub null: Option<NullSettings>,
    pub tcp: Option<TcpSettings>,
}

#[derive(serde::Deserialize, Clone)]
pub struct NullSettings {
    pub name: String,
}

#[derive(serde::Deserialize, Clone)]
pub struct UnixSettings {
    pub name: String,
}

#[derive(serde::Deserialize, Clone)]
pub struct TcpSettings {
    pub address: String,
}
```

I believe above is valid yaml (not the best looking yaml but I'll add more things to the hashtable later so the structure is needed. This worked with 0.13.x

goingforbrooke:
Confirmed. Thanks for the Issue @mjaakkol! 🙏🏻

Seems like the mixture of string and integer in `"localhost:1234" is unexpected.

Backtrace:

```
   3: config::file::format::yaml::from_yaml_value
             at ./src/file/format/yaml.rs:53:26

```

Assumed `SerialSettings` is:

```
#[derive(serde::Deserialize, Clone)]
pub struct SerialSettings {
    pub name: String,
}
```

goingforbrooke:
Nevermind-- the underlying YAML library is "guessing" that `localhost:1234` is a hash. I'll look into making ours "guess" better.

```rust
    match *value {
...
yaml::Yaml::Integer(value) => Ok(Value::new(uri, ValueKind::I64(value))),
        yaml::Yaml::Boolean(value) => Ok(Value::new(uri, ValueKind::Boolean(value))),
        yaml::Yaml::Hash(ref table) => {
            let mut m = Map::new();
            for (key, value) in table {
                match key {
                    yaml::Yaml::String(k) => m.insert(k.to_owned(), from_yaml_value(uri, value)?),
[...]

PR #650, opened 2025-03-10
feat: Add support for float and boolean hash keys

MartinHowarthFlexciton:
The `unreachable!` code at https://github.com/rust-cli/config-rs/blob/main/src/file/format/yaml.rs#L53 is reachable, and this PR fixes it for two simple types: reals and booleans.

Prior to this change you'd get this error trying to parse the following:
```
thread '...' panicked at .../config-0.15.9/src/file/format/yaml.rs:53:26:
internal error: entered unreachable code
```

The problematic case for me turned out to be float keys:
```yaml
inner_float:
    0.1: "float 0.1"
    0.2: "float 0.2"
```

The change in this PR follows the same pattern as already exists for Strings and Integers - just parse them as strings and leave it up to the user to work out what to do with that.

Personally I only need the Real support, so happy to remove the Boolean case if that's at all contentious.

I decided not to look into whether you can craft a hash with the other Yaml types (below) because it's such an odd use case it didn't seam realistic! That said, https://github.com/rust-cli/config-rs/issues/547 suggests that just treating Hash as string would have avoided that users' problem. However, converting from these types back into the exact string format they were parsed in would be difficult, so it's not clear what the right implementation would be. So getting the simpler case (reals, booleans) done feels like the right sized change to make.
```
    Array(Array),
    Hash(Hash),
    Alias(usize),
    Null,
    BadValue,
 ```

Let me know what you think! And thanks for maintaining this useful crate :)

### DIFF
--- a/src/file/format/yaml.rs
+++ b/src/file/format/yaml.rs
@@ -51,6 +51,7 @@ fn from_yaml_value(
                     yaml::Yaml::String(k) => m.insert(k.to_owned(), from_yaml_value(uri, value)?),
                     yaml::Yaml::Integer(k) => m.insert(k.to_string(), from_yaml_value(uri, value)?),
                     yaml::Yaml::Boolean(k) => m.insert(k.to_string(), from_yaml_value(uri, value)?),
+                    yaml::Yaml::Real(k) => m.insert(k.to_owned(), from_yaml_value(uri, value)?),
                     other => Err(Box::new(UnsupportedHashKeyError(format!("{other:?}"))))?,
                 };
             }
--- a/tests/testsuite/file_yaml.rs
+++ b/tests/testsuite/file_yaml.rs
@@ -382,3 +382,22 @@ inner_bool:
     assert_eq!(config.inner_bool.get(&true).unwrap(), "bool true");
     assert_eq!(config.inner_bool.get(&false).unwrap(), "bool false");
 }
+
+#[test]
+fn test_yaml_parsing_float_hash() {
+    let result = Config::builder()
+        .add_source(File::from_str(
+            r#"
+inner_float:
+    0.1: "float 0.1"
+    0.2: "float 0.2"
+"#,
+            FileFormat::Yaml,
+        ))
+        .build();
+    assert!(result.is_err());
+    assert_data_eq!(
+        result.unwrap_err().to_string(),
+        str!["Cannot parse Real(\"0.1\") because it is an unsupported hash key type"]
+    );
+}
--- a/tests/testsuite/file_yaml.rs
+++ b/tests/testsuite/file_yaml.rs
@@ -385,7 +385,12 @@ inner_bool:
 
 #[test]
 fn test_yaml_parsing_float_hash() {
-    let result = Config::builder()
+    #[derive(Debug, Deserialize)]
+    struct TestStruct {
+        inner_float: HashMap<String, String>,
+    }
+
+    let config = Config::builder()
         .add_source(File::from_str(
             r#"
 inner_float:
@@ -394,10 +399,10 @@ inner_float:
 "#,
             FileFormat::Yaml,
         ))
-        .build();
-    assert!(result.is_err());
-    assert_data_eq!(
-        result.unwrap_err().to_string(),
-        str!["Cannot parse Real(\"0.1\") because it is an unsupported hash key type"]
-    );
+        .build()
+        .unwrap()
+        .try_deserialize::<TestStruct>()
+        .unwrap();
+    assert_eq!(config.inner_float.get("0.1").unwrap(), "float 0.1");
+    assert_eq!(config.inner_float.get("0.2").unwrap(), "float 0.2");
 }

### CODE REVIEW

**Comment by epage** [comment]
Seems we should turn that panic into an error

File: src/file/format/yaml.rs [hunk 1]
```diff
@@ -50,6 +50,8 @@ fn from_yaml_value(
                 match key {
                     yaml::Yaml::String(k) => m.insert(k.to_owned(), from_yaml_value(uri, value)?),
                     yaml::Yaml::Integer(k) => m.insert(k.to_string(), from_yaml_value(uri, value)?),
+                    yaml::Yaml::Real(k) => m.insert(k.to_string(), from_yaml_value(uri, value)?),
```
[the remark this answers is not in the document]
MartinHowarthFlexciton: Good point, do you have a suggestion on how to handle that? Is there a "pass through unchanged" option we can convince the underlying yaml parser to do? 

Alternatively, is supporting float keys in a way that's not roundtrippable better than not supporting them at all? The use case of assuming the user will parse the resulting string into a float should be close to what they would expect (within some float precision error) even if the string ends up being different. We'd have to document it.

MartinHowarthFlexciton:
I've turned the remaining unsupported cases into a useful error message as you suggested

tests/testsuite/file_yaml.rs
```diff
@@ -338,3 +338,71 @@ fn yaml() {
     let date: DateTime<Utc> = s.get("yaml_datetime").unwrap();
     assert_eq!(date, Utc.with_ymd_and_hms(2017, 6, 12, 10, 58, 30).unwrap());
 }
+
+#[test]
+fn test_yaml_parsing_float_hash() {
```
epage, line 343: Sorry, I forgot to call out that we generally ask for tests to be added in their own commit before behavior changes, passing.  This shows the current behavior (e.g. `should_panic`).  Then in the behavior-change commit, the diff of the test shows how behavior changed.  This helps make sure we are testing the right thing and is one of the clearest ways to demonstrate the intent of a PR.

src/file/format/yaml.rs
```diff
@@ -50,7 +50,9 @@ fn from_yaml_value(
                 match key {
                     yaml::Yaml::String(k) => m.insert(k.to_owned(), from_yaml_value(uri, value)?),
                     yaml::Yaml::Integer(k) => m.insert(k.to_string(), from_yaml_value(uri, value)?),
-                    _ => unreachable!(),
+                    yaml::Yaml::Real(k) => m.insert(k.to_string(), from_yaml_value(uri, value)?),
+                    yaml::Yaml::Boolean(k) => m.insert(k.to_string(), from_yaml_value(uri, value)?),
+                    other => Err(Box::new(UnsupportedHashKeyError(format!("{other:?}"))))?,
```
epage, line 55: Ideally the error change would be in its own commit

tests/testsuite/file_yaml.rs
epage: (sorry, forgot to call that out earlier)

src/file/format/yaml.rs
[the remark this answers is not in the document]
epage: > Alternatively, is supporting float keys in a way that's not roundtrippable better than not supporting them at all?

That depends.  If we make a feature that will randomly break on people in unpredictable ways, is it better to not have the feature at all?  That is what we need to better understand.

> The use case of assuming the user will parse the resulting string into a float should be close to what they would expect (within some float precision error) even if the string ends up being different. We'd have to document it.

My concern is someone puts `0.1` in their yaml file and `"0.1"` in their code and expect it to work but instead the key is `"0.1000000000001"`.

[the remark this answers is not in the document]
epage: btw we could separate out these questions and merge the bool and panic->error parts of this now.

tests/testsuite/file_yaml.rs
MartinHowarthFlexciton: No worries, I clearly didn't read the contributing doc closely enough :)

src/file/format/yaml.rs
MartinHowarthFlexciton: See https://github.com/rust-cli/config-rs/pull/651

src/file/format/yaml.rs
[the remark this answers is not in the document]
MartinHowarthFlexciton: I've done that, see https://github.com/rust-cli/config-rs/pull/651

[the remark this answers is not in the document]
MartinHowarthFlexciton: So I think we don't need to worry about float imprecision being introduced within this crate. If we look at the definition of the underlying Yaml parser we see this:
```
pub enum Yaml {
    /// Float types are stored as String and parsed on demand.
    /// Note that `f64` does NOT implement Eq trait and can NOT be stored in `BTreeMap`.
    Real(String),
    <snip>
}
```
Which shows that Reals are stored as Strings leaving parsing to float as an exercise for the user, or on demand somewhere further down the yaml stack. Crucially that means the yaml crate hasn't already introduced any float imprecision.

Using that, the hash parsing in this crate initially stores all hash keys as strings, and later parsing might try to convert it to a different type. Following that chain, when using this crate, if a user tries to do the obvious approach to parse into floats:
```
    #[derive(Debug, Deserialize)]
    struct TestStruct {
        inner_float: HashMap<f64, String>,
    }
```
They will get an error saying `the trait bound `f64: std::cmp::Eq` is not satisfied`. Their only viable route forwards is either to parse as a String and then handle conversion to f64 themselves if needed; or to write a custom parser which ends up meaning the same thing.

Based on that, I think it's safe/sensible for this crate to handle yaml float hash keys. As far as I can tell we don't (and the underlying yaml crate doesn't) convert to float and back again so we're not at risk of having the string "0.1" turn up as "0.10000001". For comparison: the integer (and boolean) parsing _does_ convert to i64 and back to string.
It is possible the user makes that mistake in their custom handling of floats however. I'm not sure what this crate could/should do about that though.

What do you think?

P.S. I created a separate PR for the boolean and error messages as you suggested and I've rebased this PR on top of it - the first 4 commits will disappear if you want to merge the other one first. Up to you how you want to handle this.

src/file/format/yaml.rs
```diff
@@ -51,6 +51,7 @@ fn from_yaml_value(
                     yaml::Yaml::String(k) => m.insert(k.to_owned(), from_yaml_value(uri, value)?),
                     yaml::Yaml::Integer(k) => m.insert(k.to_string(), from_yaml_value(uri, value)?),
                     yaml::Yaml::Boolean(k) => m.insert(k.to_string(), from_yaml_value(uri, value)?),
+                    yaml::Yaml::Real(k) => m.insert(k.to_string(), from_yaml_value(uri, value)?),
```
epage, line 54: If `k` is a string, then we should call `to_owned` on it

src/file/format/yaml.rs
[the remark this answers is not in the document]
epage: If its stored as a `String`, then that works!

src/file/format/yaml.rs
MartinHowarthFlexciton: Done

PR #673, opened 2025-07-25
Bumped a few crates, ran cargo update

martinfrances107:
-ron = { version = "0.8", optional = true }
+ron = { version = "0.10", optional = true }

-convert_case = { version = "0.6", optional = true } 
+convert_case = { version = "0.8", optional = true }

-notify = "7.0"
+notify = "8.1"

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -5,7 +5,7 @@ resolver = "2"
 repository = "https://github.com/rust-cli/config-rs"
 license = "MIT OR Apache-2.0"
 edition = "2018"
-rust-version = "1.75.0"  # MSRV
+rust-version = "1.75.0" # MSRV
 include = [
   "build.rs",
   "src/**/*",
@@ -13,7 +13,7 @@ include = [
   "Cargo.lock",
   "LICENSE*",
   "README.md",
-  "examples/**/*"
+  "examples/**/*",
 ]
 
 [workspace.lints.rust]
@@ -55,7 +55,7 @@ invalid_upcast_comparisons = "warn"
 large_digit_groups = "warn"
 large_stack_arrays = "warn"
 large_types_passed_by_value = "warn"
-let_and_return = "allow"  # sometimes good to name what you are returning
+let_and_return = "allow"                 # sometimes good to name what you are returning
 linkedlist = "warn"
 lossy_float_literal = "warn"
 macro_use_imports = "warn"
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -5,15 +5,15 @@ resolver = "2"
...
+rust-version = "1.75.0" # MSRV
 include = [
   "build.rs",
   "src/**/*",
   "Cargo.toml",
   "Cargo.lock",
   "LICENSE*",
   "README.md",
-  "examples/**/*"
+  "examples/**/*",
```
epage, line 16: There seem to be unrelated changes

Cargo.toml
```diff
@@ -113,50 +113,73 @@ rustdoc-args = ["--cfg", "docsrs", "--generate-link-to-definition"]
...
+] }
 serde_json = { version = "1.0", optional = true }
 yaml-rust2 = { version = "0.10", optional = true }
 rust-ini = { version = "0.21", optional = true }
-ron = { version = "0.8", optional = true }
+ron = { version = "0.10", optional = true }
 json5_rs = { version = "0.4", optional = true, package = "json5" }
 indexmap = { version = "2.10.0", features = ["serde"], optional = true }
-convert_case = { version = "0.6", optional = true }
+convert_case = { version = "0.8", optional = true }
```
epage, line 162: This is a breaking change and we'd need to handle this more conciously

Cargo.toml
```diff
@@ -113,50 +113,73 @@ rustdoc-args = ["--cfg", "docsrs", "--generate-link-to-definition"]
...
-toml = { version = "0.9", optional = true, default-features = false, features = ["parse", "serde"] }
+toml = { version = "0.9", optional = true, default-features = false, features = [
+  "parse",
+  "serde",
+] }
 serde_json = { version = "1.0", optional = true }
 yaml-rust2 = { version = "0.10", optional = true }
 rust-ini = { version = "0.21", optional = true }
-ron = { version = "0.8", optional = true }
+ron = { version = "0.10", optional = true }
```
epage, line 159: We are not ready to upgrade ron, see https://github.com/rust-cli/config-rs/pull/476#pullrequestreview-2698875901

Cargo.toml
```diff
@@ -113,50 +113,73 @@ rustdoc-args = ["--cfg", "docsrs", "--generate-link-to-definition"]
...
+  "io-util",
+  "time",
+] }
 warp = "0.3"
 futures = "0.3"
 reqwest = "0.12"
 
 glob = "0.3"
-notify = "7.0"
+notify = "8.1"
```
epage, line 182: This requires an MSRV bump.  The question is whether its worth it for a dev dependency

alexanderkjall: isn't it possible to have a higher MSRV for dev-deps? glob seem to have it: https://github.com/rust-lang/glob/blob/master/.github/workflows/rust.yml

epage: Yes, but then you can't run tests and verify the behavior under MSRV, see also https://doc.rust-lang.org/cargo/reference/rust-version.html#support-expectations

PR #677, opened 2025-08-12
perf(json5): Remove use of serde(untagged)

epage:
Prep for #472

### DIFF
--- a/src/file/format/json5.rs
+++ b/src/file/format/json5.rs
@@ -4,8 +4,7 @@ use crate::format;
 use crate::map::Map;
 use crate::value::{Value, ValueKind};
 
-#[derive(serde::Deserialize, Debug)]
-#[serde(untagged)]
+#[derive(Debug)]
 pub(crate) enum Val {
     Null,
     Boolean(bool),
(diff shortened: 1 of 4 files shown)


src/file/format/json5.rs
```diff
@@ -16,6 +15,23 @@ pub(crate) enum Val {
...
+        D: serde::de::Deserializer<'de>,
+    {
+        serde_untagged::UntaggedEnumVisitor::new()
+            .bool(|value| Ok(Self::Boolean(value)))
+            .i64(|value| Ok(Self::Integer(value)))
+            .f64(|value| Ok(Self::Float(value)))
+            .string(|value| Ok(Val::String(value.to_owned())))
+            .none(|| Ok(Self::Null))
+            .seq(|value| value.deserialize().map(Val::Array))
+            .map(|value| value.deserialize().map(Val::Object))
```
dtolnay, line 30: Since this PR is partially motivated by performance, it is worth knowing about https://github.com/dtolnay/erased-serde/issues/61 (which is used by serde-untagged). For very deeply nested config you would see O(n²) performance in the new implementation compared to O(n) in the previous one. Ultimately the best performance would be from a more verbose impl that resembles https://github.com/serde-rs/json/blob/v1.0.142/src/value/de.rs#L22, which completely eliminates the quadratic factor and also reduces linear factor from heap allocations inside erased-serde and constant factor from heap allocations inside UntaggedEnumVisitor.

PR #679, opened 2025-08-15
Update build status badge

atouchet:


epage:
Thanks for pointing this out!  It turns out that this was not matching our template, so I fixed that in 04d6dee.

PR #682, opened 2025-08-20
chore: Remove `add_dummy_extension()` helper

polarathene:
The helper method is redundant (_introduced as a [bug fix in Mar 2022](https://github.com/rust-cli/config-rs/pull/306)_), there is a much simpler approach to append the extension.

- [`as_mut_os_string()`](https://doc.rust-lang.org/std/path/struct.Path.html#method.as_mut_os_str) has been available since Rust `1.70.0`, matching the MSRV of this crate since the introduction of [`OnceLock`](https://doc.rust-lang.org/std/sync/struct.OnceLock.html) [in Oct 2024](https://github.com/rust-cli/config-rs/pull/515) (_`0.14.1` release_).
- Additionally revised the comment and included a conditional statement to better communicate the intent for why this placeholder is necessary.

### DIFF
--- a/src/file/source/file.rs
+++ b/src/file/source/file.rs
@@ -59,9 +59,13 @@ impl FileSourceFile {
                 )))
             };
         }
-        // Adding a dummy extension will make sure we will not override secondary extensions, i.e. "file.local"
-        // This will make the following set_extension function calls to append the extension.
-        let mut filename = add_dummy_extension(filename);
+
+        let mut filename = filename;
+        // Preserve any extension-like text within the provided file stem by appending a fake extension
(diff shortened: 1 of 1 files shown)


polarathene:
Lint failure is about commit message line length of 72.

I personally consider that legacy, most interfaces for viewing commit messages should have the capability to line wrap 🤷‍♂️

That said, in the actual PR I did split the comment line to be under 120 characters in length. UX is harmed a tad when line length of the viewer is below that if line wrapping is enabled 😅 Let me know if you'd rather it be folded back into a single line (207 length).

epage:
> Lint failure is about commit message line length of 72.

As I said elsewhere and in the CI job message, the lint commit job is more of a suggestion than a hard requirement.

epage (approved):
Thanks!

PR #683, opened 2025-08-21
feat(env): add source_os method for safe Unicode env handling

gustavovalverde:
## Problem
The current `Environment::source()` method has critical limitations:
1. **Unicode Panics**: Using `env::vars()` panics on invalid Unicode
2. **Race Conditions**: TOCTOU vulnerability when filtering environment variables  
3. **API Limitation**: Only supports `Map<String, String>`, blocking OS path handling

This was originally identified by @str4d

More context: https://github.com/zcash/wallet/pull/194/files#r2288154972

## Solution
Add `source_os()` method that accepts `Map<OsString, OsString>`:

- Uses `env::vars_os()` internally for safe Unicode handling
- Eliminates race conditions with atomic snapshots
- Maintains full backward compatibility
- Takes precedence over `source()` when both are set

## Example Usage
```rust
// Safe filtering without Unicode panics or race conditions
let mut filtered_env = Map::new();
for (key, value) in env::vars_os() {
    if let Some(key_str) = key.to_str() {
        if key_str.starts_with("PREFIX_") && !is_sensitive(key_str) {
            filtered_env.insert(key, value);
        }
    }
}

Environment::with_prefix("PREFIX").source_os(Some(filtered_env))
```

## Tests
- All existing tests pass (148/148)
-  New Unicode safety tests added
- Precedence behavior tested
- Example with real-world usage included

### DIFF
--- a/examples/safe_env_filtering/main.rs
+++ b/examples/safe_env_filtering/main.rs
@@ -1,0 +1,93 @@
+use std::ffi::OsString;
+
+use config::{Config, Environment, Map};
+use serde_derive::Deserialize;
+
+#[derive(Debug, Deserialize)]
+struct AppConfig {
+    debug: bool,
+    port: u16,
+    secret_key: Option<String>, // This should NOT be overridable via env
+}
+
+/// Example of safe environment variable filtering
+///
+/// This demonstrates how to safely filter environment variables
+/// without risking Unicode panics or race conditions.
+fn main() -> Result<(), Box<dyn std::error::Error>> {
+    // Simulate setting some environment variables
+    std::env::set_var("MYAPP_DEBUG", "true");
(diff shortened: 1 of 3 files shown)


epage:
Looking over the points of the PR, I'm not seeing a reason to continue forward with this.

If there is something I missed, please open an Issue per distinct concern as our [contributing guide](https://github.com/rust-cli/config-rs/blob/main/CONTRIBUTING.md#pull-requests) asks.

> Unicode Panics: Using env::vars() panics on invalid Unicode

No, it doesn't

> Race Conditions: TOCTOU vulnerability when filtering environment variables

What is the justification for this?

> API Limitation: Only supports Map<String, String>, blocking OS path handling

This whole library has this limitation

PR #684, opened 2025-08-27
refactor: `find_file()` should be easier to grok

polarathene:
I felt the current method was a bit of a mess and could be made easier to digest by cleaning up the control flow.

### Context

When I glanced over this method I found it tricky to follow, especially with the use of `extensions()` for `FileFormat` fallbacks instead of sharing the common `file_extensions()` it also supports.

I've gone through quite a few iterations to land on this one. It primarily focuses on a clean control flow for a high-level noise-free function to grok.

I recall having some trouble / complexity from trying to include `file_exists_with_format()` within `find_file()`, and keeping this DRY as opposed to the present duplication prior to the PR was a goal. `impl Trait` with the generic `F: FileStoredFormat` keeps the functionality leveraging static dispatch with the two different types, and extracting this out to a function provides that clearer input/output signature while keeping the main function body focused.

I inverted the control flow between `filename` vs `format_hint` conditionals, which was helpful due to the more functional approach since I could then unify `format_hint` and `FileFormat` branches when casting their type to `dyn Format`. I also find it easier to reason the distinction of the two types with their fallbacks that try filenames with extensions (_vs interleaving the handling between the two types based on the original `filename.is_file()` as the first condition_).

I've included some additional comments where I thought it might be helpful context, notably for the match statement in `find_file()` it may provide some minor value vs parsing the whole thing to piece that together? 😅 Open to trimming it out if it doesn't seem of value (_seems to align with the intent of comments in the other function at the bottom of the file that I've not touched_).

---

### Change overview

- Now more consistent across branches
- Errors extracted out to functions
- Easier control flow to grok:
  - Start with `format_hint`
  - Fallback to `FileFormat`
  - Fail early with `ok_or_else()?`


src/file/source/file.rs
```diff
@@ -27,77 +27,83 @@ impl FileSourceFile {
...
-        Err(Box::new(io::Error::new(
-            io::ErrorKind::NotFound,
-            format!(
-                "configuration file \"{}\" not found",
-                self.name.to_string_lossy()
-            ),
-        )))
+        Box::new(io::Error::new(io::ErrorKind::NotFound, error_message))
+    }
+}
```
polarathene, line 87: Not entirely a fan of these functions but they're presumably fine being handled like this? I don't think there's much value embedding them into `find_file()`, and since I swapped the first conditional to `format_hint` matching, the `error_invalid_path` would need to be repeated.

I'm not sure if it's relevant, but `error_invalid_format` does differ slightly from the original error definition which used `filename` instead of `self.name` for the message param. I think the intent is to remain consistent with the two errors, so I normalized on `self.name`.

---

Perhaps the crate could bring in a dep like `thiserror` and we'd have something like this instead?:

```rust
#[derive(Debug, thiserror::Error)]
enum ConfigFileError {
  #[error("Failed to identify a supported config format for: '{}'", .0.to_string_lossy())]
  UnsupportedFormat(PathBuf),

  #[error("Configuration file '{}' not found", .0.to_string_lossy())]
  NotFound(PathBuf),
}
```

Then it's just something like `.ok_or_else(|| ConfigFileError::UnsupportedFormat(self.name))`?

---

When updating `Cargo.lock` I noticed it implicitly had `thiserror` deps:
- v1 (dev-dep: `warp@0.3` via `tokio-tungstenite` for websockets, changed from default to opt-in in `warp@0.4`)
- v2 (feature: `json5` via `pest`)

src/file/source/file.rs
```diff
@@ -27,77 +27,83 @@ impl FileSourceFile {
...
+
+fn identify_format<F: FileStoredFormat>(filename: &Path) -> impl FnMut(&F) -> bool + '_ {
+    let ext = filename.extension().unwrap_or_default().to_string_lossy();
+    move |format| format.file_extensions().contains(&ext.as_ref())
+}
```
polarathene: This doesn't need to be split out into a function, given that it's just two lines in the body, but it does has a slight benefit of co-location to `file_Exists_with_format()` that follows it, along with a clearer signature (_and at the call-site it's a bit neater of an overview with the function reference vs inlining into the branch_).

If it were better to be inlined for the reader, I would suggest adjusting it's branch to this:

```rust
let identify_format = |ext: &std::ffi::OsStr| {
    let ext = ext.to_str()?;
    internal_formats.find(|f| f.file_extensions().contains(&ext))
};

filename
	.extension()
	.and_then(identify_format)
	.ok_or_else(error_invalid_format)?
```

That way the `unwrap_or_default()` equivalent would bail early as `None`, which is a minor optimization to skip iterating through formats to find a match. It also makes usage of `ext` a little nicer. The above has been put through RustFmt for good measure 😅 (_since I'm familiar with it's habit to split a perfectly fine one-liner into many lines, which a few variations of the above code doesn't look as pleasant afterwards_)

polarathene:
The MSRV CI job fails due to the use of `use<'_, F>` in the return type here:

```rust
fn file_exists_with_format<F: FileStoredFormat>(filename: &mut PathBuf) -> impl FnMut(&F) -> bool + use<'_, F> {
```

```rust
   error: expected identifier, found keyword `use`
    --> src/file/source/file.rs:95:101
     |
  95 | fn file_exists_with_format<F: FileStoredFormat>(filename: &mut PathBuf) -> impl FnMut(&F) -> bool + use<'_, F> {
     |                                                                                                     ^^^ expected identifier, found keyword
  
  error[E0405]: cannot find trait `r#use` in this scope
    --> src/file/source/file.rs:95:101
     |
  95 | fn file_exists_with_format<F: FileStoredFormat>(filename: &mut PathBuf) -> impl FnMut(&F) -> bool + use<'_, F> {
     |                                                                                                     ^^^ not found in this scope
```

In Rust 2024 edition this is not necessary anymore, but it was a complaint from the compiler to add it when I compiled for an earlier edition. I'll compile with Rust 1.75, assuming that `use` is dependent upon a newer version of Rust or that something else is up 🤷‍♂️

I believe I can also get away with `+ '_` instead of `+ <'_, F>` 🤔

epage:
With a refactor like this, can you split it out into the individual refactor steps (extracting a function, renaming a variable, etc)?  It would make it easier to follow whats happening

src/file/source/file.rs
epage: > I'm not sure if it's relevant, but error_invalid_format does differ slightly from the original error definition which used filename instead of self.name for the message param.

Please extract changes of behavior into their own commit

src/file/source/file.rs
```diff
@@ -27,77 +27,83 @@ impl FileSourceFile {
...
+
+// Provides a predicate to verify a file exists with an extension that is compatible with the queried format.
+// To simplify usage at the call sites, `filename` will be mutated by reference with each use of the closure.
+fn file_exists_with_format<F: FileStoredFormat>(filename: &mut PathBuf) -> impl FnMut(&F) -> bool + use<'_, F> {
```
epage: The name for this function does communicate how it mutats `filename` and I feel combining filtering with mutation to be brittle and non-obvious. In general, I feel that iterator combinators should not do mutations.

polarathene: While I tend to agree with you, there is a clear contextual comment above, the functional signature also communicates this, and at the call-site it is also apparent due to passing in `&mut filename` into the call. So the intent should be rather obvious at this boundary?

I had iterated on the function name multiple times but this was the best one that I could come up with for it's purpose, leaning into the function signature and visible `&mut` at the call site to communicate the filename being mutated via reference. Rather than bike shed it too much I figured that was sufficient for a private function that barely anyone is going to read going forward, but the goal of making the whole `find_file()` flow itself easier to grok was still accomplished.

Prior to this the mutations were happening within the `find_file()` body, and all you had to go on was `let mut filename` and the `set_extension()` calls. Thus arguably this is more explicit about the source of the mutation being at the function for a mutable variable that was implicitly modified via `.set_extension()` calls?

I find with Rust this concern with side-effects is less problematic thanks to the type system more clearly communicating this information for us, whereas with other languages I've worked with it's much more ambiguous.

polarathene: > I feel combining filtering with mutation to be brittle and non-obvious. In general, I feel that iterator combinators should not do mutations.

Personally I find it much more readable and acceptable. The iterators themselves are using a predicate closure with the closure generator here capturing the input for a mutable reference being very clear.

In the case of the `filter()` call (_which is on an option not an iterator_), this even reads well with the function name IMO:

```rust
let valid_format = format_hint
    .filter(|format| {
        filename.is_file() || file_exists_with_format(&mut filename)(format)
    })
    .ok_or_else(|| self.error_invalid_path())?;

Ok((filename, Box::new(valid_format)))
```

Take an option and retain a `Some(_)` value if:
- The filename exists already (short-circuit)
- Or the filename exists with a format compatible extension (_ext was left out of the fn name to not confuse the ext as input_)
- Otherwise the path is invalid (_we already know it was originally `Some(_)` from the start, the `.filter()` discards it only if the filename cannot be identified_)

Contrast this to the two split `match` statements of the prior existing approach, each in different conditional branches where the alternative `match` arms are for the internal fallback type `FileFormat`, with eerily similar looking code duplication (_but calls to `.extension()` or `file_extension()`_) and nesting... The above is much more terse and readable.

From that snippet alone, without seeing the body of `file_exists_with_format`, it is clearly communicated that:
- We've taken a `format_hint` (_that has a value due to the `match` arm predicate it's associated to_)
- We are concerned about the filename being found to retain the format hinted.
- After all that we return the expected successful result of the filename and boxed format hint.

If the filename was not to mutate, that `&mut filename` at the call site would be pointless, and filename would have no reason to be mutable.

---

For comparision here is the original equivalent logic I've replaced:

```rust
if filename.is_file() {
    return if let Some(format) = format_hint {
        Ok((filename, Box::new(format)))
    } else {
        // ... fallback logic for `FileFormat`
        Err( ... ) 
    };
}

// Implicit else
let mut filename = filename;
if filename.extension().is_some() {
    filename.as_mut_os_string().push(".placeholder");
}

match format_hint {
    Some(format) => {
        for ext in format.file_extensions() {
            filename.set_extension(ext);

            if filename.is_file() {
                return Ok((filename, Box::new(format)));
            }
        }
    }
    // The fallback has roughly the same logic here as for the Some() arm above
    None => { ... }
}

Err( ... )
```

I condensed and localized that mess for `format_hint` into the few lines I showed earlier + the separate extracted function call.

We have the `let mut filename` at the start of the implicit else branch, with a divergence at the following `match` as both match arms rely upon mutating `filename` with roughly the same logic to use `.set_extension()`. I'm not seeing how that is cleaner/preferable vs a more explicit side-effect relation communicated like how I've demonstrated with my approach?

I've retained the same `Ok()` statement:
- It's better this way for the implicit coercion into the return type, which otherwise requires a hint if transforming via `.map()` (_which due to ownership concerns also required a 2nd map to get the tuple with `filename`, or introduce a `clone()` call_).
- Albeit without an explicit `return` as my approach does not require that. `filename` must be inferred as mutated from earlier context, or after this `match` the fallback to returning an `Err()`.

Which of the two is easier to read and understand without the verbose breakdown?

You get an added bonus of not only different types (`F` vs `FileFormat`) but:
- One works on owned values while the other via references (_for which you can run into some concerns with traits not implemented for referenced types_), and these types differ in capabilities (_one does not support `Copy`/`Clone` for example_).
- Which I believe was originally why `FileFormat` was calling `.extension()` instead of `.file_extenions()` (_at least this was a problem I noticed when initially adapting for consistency_), and so if you're not familiar with the codebase, you need to go hunt down those traits/logic in other files to grok the differences.

---

By normalizing both match arms to use the same logic, complexity was reduced, and the `file_exists_with_format()` method to generate closures leverages `impl Trait` with generics to avoid any implied compatibility of the two different types sharing the same closure (_unless leveraging dynamic dispatch which I chose to avoid_).
- That only works with a separate method as I've introduced here since closures do not support `impl Trait` (or generics IIRC), and you'd run into some ownership issues if you try to capture `filename` into a closure at an earlier stage within `find_file()`.
- This approach I've taken has less complexity to support it and it introduces a clear boundary for the common logic and explicit mutable reference.

You could add some minor overhead and clone the PathBuf, but then you have to handle `filename` being sourced from multiple points instead of a common mutable variable all branches source from. Which for a small and simple function body as this PR proposes, I would deem this quite reasonable / acceptable.

The control flow is vastly simpler:

```rust
let mut filename = ...;

match format_hint {
    Some(_) | None => {
        // This generated predicate will mutate filename per call:
        let mut generated_predicate = file_exists_with_format(&mut filename);

        // ... actual logic per match arm here
        // if not `filename.is_file()` use `generated_predicate(&format)`

        // Regardless upon success return `filename from the same original mutable variable
        Ok((filename, Box::new(valid_format)))
   }
}
```

There is so much more noise and nesting going on with `filename` interactions in the original `find_file()` prior to this PR, along with a rebinding midway that it's not as simple to reason about a simplified control flow like the pseudo-code representation above depicts.

polarathene: Apologies for the verbosity above, I just wanted to be very clear about my reasoning for why this was appropriate in this scenario.

It has clear benefits, I do tend to agree with you that side-effects should be avoided when possible but I did not see a nicer way to approach this without introducing extra overheads, and in previous communications you've not been fond of regressing performance to benefit DX.

Perhaps like the last snippet above shows, you'd be comfortable with an assignment of the closure to a `mut` variable? (_with optional contextual comment_). I didn't feel that was adding much value personally, I wanted to keep things DRY and clear but maybe it doesn't read like that to fresh eyes 😅 

---

Should you have a name for this method that you would find more suitable, please feel free to suggest it and I can change it.

If you're still strongly against the approach I've taken here and would like to suggest an alternative, feel free to. I spent quite a bit of time fiddling around with this refactoring to land on this revision for a PR, I hope you can see how much easier it is to digest `find_file()` and that the above alleviates your concerns raised with `filename`.

epage: I missed the mutation.

polarathene:
> With a refactor like this, can you split it out into the individual refactor steps (extracting a function, renaming a variable, etc)? It would make it easier to follow whats happening

I had considered this, but the original method was such a tangled mess with my refactor quite the departure (_from nested loops/conditionals/returns to combinators + functions_), given the scope of the change I wasn't sure if it was worth the additional effort.

- I provided a [detailed before/after comparison in another review feedback comment](https://github.com/rust-cli/config-rs/pull/684#discussion_r2305768747) which should explain the changes far better?
- Otherwise it's a bit tricky for how you would want this single method in a single file having the refactor split into meaningful commits? (_`CONTRIBUTING.md` also mandates that each commit can be built successfully, not solely to provide a meaningful diff, which is another perk of squash merges for PRs btw_ 😉)

For the most part, apart from switching to combinators and extracting logic that is easy to match to their functions, nothing has changed that much in what is done. No variables renamed (_other than changing `filename` variable to `mut` earlier due to condensing the conditional branching_).

I'll go ahead and split it (_it's kind of nice, when I did make such efforts in contributions to other projects, separate diffs and commit messages rarely seemed appreciated_ 😅)

---

I could make the following commits if you'd like:
- Extract out the errors to their individual function calls.
  - **EDIT:** With a separate commit apparently for [one small change](https://github.com/rust-cli/config-rs/pull/684#discussion_r2304185657) (_when `filename.is_file()` + no `format_hint`, the absolute path for `filename` was used instead of the structs `self.name` which the other error uses_).
- Replace the match arm for loops with equivalent changes this PR introduces, keeping any duplication. I'm not 100% sure if that compiles correctly off the top of my head. It'll potentially need some temporary helper methods due to some differences.
- Extract out that replaced logic to the `identify_format()` + `file_exists_with_format()` methods.
- Now I could invert the `is_file()` + `match` conditionals to match my PRs single `match` with the two different format types logic unified instead of interleaved.

So mostly just shuffling data around and some extra temporary noise to improve the diff readability, I'm all for that and will give it a shot but if a commit introduces too much friction to meet the build + tests passing requirement I will opt to flatten rather than sink too much time into massaging a readable diff.

src/file/source/file.rs
polarathene: > I missed the mutation.

I'm not sure how to interpret that?
- Are you saying you didn't notice the mutation when you provided the earlier review comment about mutation side-effect concerns? (_uncertain as this feedback thread was left unresolved_)
- Or are you clarifying that despite the `&mut PathBuf` signature and function usage at the call-sites passing an explicit `&mut filename` you missed that `filename` was being mutated until coming across the function and seeing something in the function body? (`set_extension()` perhaps?)

---

The method would have to either consume the `PathBuf` or clone it to avoid the side-effect, but doing so in this case would introduce complexity and go against the goal of the PR.

I don't think I can mutate captured variables by reference without the signature explicitly communicating that, but without this separate function generating closures.. it'd definitely be easier to miss a mutation like in the code prior to this PR 😅

It's probably why the earlier Err returned used `filename` but the 2nd Err returned (_after rebinding `filename` as mutable_) preferred to use `self.name.to_string_lossy()` instead, since the mutated filename would not be ideal 🤔

polarathene:
Split out changes progressively across commits, hope the diff through those is helpful 😅 (_each commit also provides extra context in code comments or commit messages_)

Other than that just waiting on if you're now okay with `file_exists_with_format(&mut filename)` mutating the filename through a predicate closure.

src/file/source/file.rs
epage: I did not expand because the point was that it failed your assumption that the intention was obvious or easier to grok.  In fact, I was in the middle of writing a review comment about changing behavior (returning a file stem instead of a file name) before I finally noticed the mutation.   We can't excuse complexity through little helpers "that no one will read" because the point shouldn't be that you need to read them but that they should be clear from the code structure and the name as to what they do.  Similarly, if you have to rely on documenting something to solve a problem; you've already failed because people will read documentation as a last resort.

The current code structure will not be merged.

polarathene: > before I finally noticed the mutation.

Could you please explain to me how you found yourself in that scenario?
- You did not start from the call-site with a clear indication of mutation via an explicit `&mut` passed in?
- You did not start reading over the function signature itself which has a clear `&mut` signature associated to the `filename` variable that the function body clearly interacts with and could not have sourced elsewhere?

I would like to be clear that I am genuinely curious, these are not questions to belittle you (_I have much respect for you, but sometimes there is miscommunication with people when I ask questions like this_).

---

I would appreciate clarity into how the existing code prior to this PR is more acceptable. The `find_file()` body has two declarations of `filename`, one halfway through with a `mut` rebinding, and effectively the same mutation logic nested away several for loop iterations deep.

Why is that more desirable? Especially given the inconsistency in the similar branches, there is more mental overhead there to grasp the flow and mutation than my approach in this PR. I know because I tried making sense of the mess prior without the extra context the code/assurances in this PR provides (regardless of the commentary).

> Similarly, if you have to rely on documenting something to solve a problem; you've already failed because people will read documentation as a last resort.

You've created an issue on the project specifically about improving documentation of the crate 🤷‍♂️ 

The commentary isn't 100% necessary, but I don't think it's doing any harm either as some people do digest information differently than others. In this case you could manually parse the `match` statement with combinators fairly easy to get an overview, but the comment above it is complimentary.

We already had a rather verbose comment about the placeholder extension that you were quite comfortable to retain, so this feedback contrasts against that and confuses me.

I'd like to also point out existing commentary at the bottom of this file could also be seen as redundant, yet it's there and I was just respecting that the project wants to clearly communicate steps as an alternative 🤨 (_is it really that bad, or does the code appear at all reliant upon the comment?_)

https://github.com/rust-cli/config-rs/blob/00a510915056c609bfa49d643605e3fd549ec032/src/file/source/file.rs#L112-L126

When it comes to a function using `&mut` this isn't that big of an ask for the reader to be familiar with this syntax in Rust, to understand that a mutable reference has been passed in and that `filename` is sourced from that via the function parameter. I have no issue removing these contextual comments, they were intended as an additional precaution should the reader not grasp the mutation occurring for some reason, and yes I clearly failed here and I hope to understand how given this is only 4-5 lines self-contained in a function, mutation of `filename` was not apparent?

---

> The current code structure will not be merged.

As you're blocking the PR intentionally, are you going to suggest how you'd like it solved?

Do you prefer the other glaring issues I've raised with the existing version that this PR resolves? All because a private function with only two call-sites closely co-located are relying upon explicit `&mut filename`?

> We can't excuse complexity through little helpers "that no one will read" because the point shouldn't be that you need to read them but that they should be clear from the code structure and the name as to what they do

Realistically not many will read/maintain this much further, besides the slight change to `all_extensions()` 10 months ago, it's not been touched since early 2022, some of it 7-8 years old. Even if someone would engage with this area, I made the effort to make the overall functionality easier to grok than the previous iteration and added the extra _complimentary_ commentary for the benefit of said future readers.

Your concern with the `&mut filename` still applies here in the existing code no?:

```rust
let mut filename = filename;
if filename.extension().is_some() {
    filename.as_mut_os_string().push(".placeholder");
}

match format_hint {
    Some(format) => {
        for ext in format.file_extensions() {
            filename.set_extension(ext);

            if filename.is_file() {
                return Ok((filename, Box::new(format)));
            }
```

The function body is doing the equivalent of what we see above in the match arm branch, with `set_extension()` mutation and further up is `mut filename`, I really don't follow how that's vastly different than the function with a more focused / narrow body where the variable is clearly mutable via ref passed in from the function.

At the call site for both scenarios we now have a clear `.filter()` predicate:

```rust
file_exists_with_format(&mut filename)(format)
```

The file `&mut filename` should exist with a compatible extension via the provided `format`.

Likewise when used in the `.find()` call:

```rust
internal_formats
    .find(file_exists_with_format(&mut filename))
    .ok_or_else(|| self.error_invalid_path())?
```

Some expectation but `format` is implicit, since in this case we just pass the closure returned directly, like if using `.map(String::from)`, it's not an uncommon pattern. There's a clear indication that a mutable ref was involved in providing that closure, the burrowed lifetime starts/ends within `.find()`'s usage of that returned closure.

When you instead directly inline the closure is it any better from omitting that explicit `&mut` context? Is the inlining beneficial here to getting a higher-level overview of the core logic? Or does it add noise? (_especially with the repetition in this case and mutation being far less obvious_)

Here it is inlined:

```rust
if filename.extension().is_some() {
    filename.as_mut_os_string().push(".placeholder");
}

internal_formats
    .find(|format| {
        format.file_extensions().iter().any(|ext| {
            filename.set_extension(ext);
            filename.is_file()
        })
    })
    .ok_or_else(|| self.error_invalid_path())?
```

The only difference prior to this PR is that instead of being in a `.find()` it's nested within two additional `for` loops and the valid value nested in another conditional for `is_file()` to decide to bail early via `return`, error handling is not as well co-located either.

This PR provides step-by-step control flow that is very clear:
- Iterate each extension of all internal formats, find a valid format that verifies the filename exists with that extension (_mutates `filename` just as before this PR_)
- If nothing was returned bail with the desired error.
- Otherwise the format is returned when matched, the `filename` is already mutated, produce the tuple result with boxed format.

You cannot honestly say this was better prior to this PR? That was very disjoint, and ambiguous (_foreign calls `.extension()` vs `.file_extension()`, interleaved types, varied control flow paths_).

---

What about the name `file_exists_with_format(&mut filename)` is not clear in how it works when used as a predicate input?
- If you do look at the function itself, the signature indicates a closure is returned that provides a boolean result and you can see `filename` is mutated.
- With the usage the burrow checker is happy (_if you view prior commits of this PR that is not the case when it's not a direct call_), so I am rather confused what problems you're anticipating for private methods that are highly unlikely to be refactored again. The burrow checker would raise alarms and there is clear mutation to the best possible way that can be communicated (_beyond a longer redundant name_).

If you **can't be bothered to read the input or the function signature**, I highly **doubt a longer function name indicating mutation** is going to make any difference to the reader. You've already proven that, yet point it out as a concern that should be addressed 😩 All relevant locations right next to the function name have a visible `&mut filename`, it's self-documenting? (_which as you've expressed should be the goal_)

Were the requested split commits viewed? Or did I waste a good portion of my time preparing that for you? (_as it appears my responses here are due to their verbosity_) If you do view them, you will see the extra complexity that comes from shifting this logic back from the extracted function (_or we just retain a duplicate of the body as was introduced early into the commits_).

I'm trying to seek a way forward and to collaborate with your requests / preferences, but **this blocker conflicts with it's own reasoning**? 😬

---

If this function is such a deal-breaker, mutation can be shifted back into `find_file()`.

I can't imagine you'd want the alternative of a clone call per iteration, or similar allocations like via [`join()`](https://doc.rust-lang.org/std/path/struct.Path.html#method.join) with the original `filename` and extension, and return that via `find_map()` instead of `find`? I avoided such due to having another PR blocked because of performance regression concerns you raised. I'm not sure what you're expecting 😓

epage: I am not going to be reading this big walls of text justifying code that will not be merged.  Naming of the function is important but even doing mutation in what should be a pure combinator is a blocker for me.

> You did not start from the call-site with a clear indication of mutation via an explicit &mut passed in?

I did.  The mutation wasn't obvious even with `&mut`.

>  I avoided such due to having another PR blocked because of performance regression concerns you raised.

Are you referring to my performance concerns over `#[serde(untagged)]`?  That is on a different level than a couple allocations here.

polarathene:
> The current code structure will not be merged. - [Ref](https://github.com/rust-cli/config-rs/pull/684#discussion_r2307389245)

Just to clarify, the PR is blocked because of a function name not indicating mutation despite both it's signature and call-sites having an explicit `&mut` right beside the name? (_and that you do not like the mutation via `&mut` itself?_)

Just to be clear, due to the two different format types two closures would need to be generated and that can only be done via static dispatch with generics (_requires a function definition as I've provided_), dynamic dispatch, or duplication (inlining). I assumed the later two were undesirable, so a separate function made sense.

`identify_format()` does not have this concern and could be inlined. I kept it extracted to co-locate beside `file_exists_with_format()`, since if you're going to view these predicate closure creators, you're probably going to view both.

It made sense to keep the core of `find_file()` simple/clean as a logic overview, the predicates and errors distract from that given the 4 variant paths involved it seemed prudent to keep that focused (_the compiler will inline methods since Rust 1.75 so no perf concerns there AFAIK_).

---

I strongly believe this PR to cater to maintainability and readability far better than the existing code and find the reasoning for blocking to be rather puzzling. I understand you're quite busy and have better things to do, but I am at a loss for the concerns raised. View the PR file as-is, without diff or PR discussions in the way. Then look at the existing code, it should be rather self-evident...?

When you've got limited time to spare I get the impression it negatively impacts the review, in that you weren't able to notice improved clarity of mutation and cite poor code structure, as if this is not in any way an improvement over lumping all the logic into a single function body, interleaved and inconsistent, as if to imply that is somehow preferable and more maintainer friendly? 😕

epage:
> Just to clarify, the PR is blocked because of a function name not indicating mutation despite both it's signature and call-sites having an explicit &mut right beside the name? (and that you do not like the mutation via &mut itself?)

I explicitly said otherwise, see https://github.com/rust-cli/config-rs/pull/684#discussion_r2307736948

> I strongly believe this PR to cater to maintainability and readability far better than the existing code 

You understand that this has already been disproved by the fact that I missed this?  And this doesn't even get into the brittleness of relying on matching the behavior of the `bool` with the mutation.

> When you've got limited time to spare I get the impression it negatively impacts the review, in that you weren't able to notice improved clarity of mutation and cite poor code structure, as if this is not in any way an improvement over lumping all the logic into a single function body, interleaved and inconsistent, as if to imply that is somehow preferable and more maintainer friendly? 

This was not me rushing the review.

src/file/source/file.rs
polarathene: > The mutation wasn't obvious even with `&mut`.

Could you please clarify?
- Does `&mut filename` not indicate an intent for mutation to consider?
- While I agree the actual mutation doesn't stand out in the function body due to no assignment, but a `set` method call, this was the same mutation as the prior code was using already prior, just the existing code was more implicit.

> doing mutation in what should be a pure combinator is a blocker for me.

While I disagree with this being a problem in Rust, I understand the concern and I'll address it 👍

> Are you referring to my performance concerns over `#[serde(untagged)]`?

Yes. Your concerns there wasn't likely to be applicable to the vast majority of real-world usage I think (_although quite valid for when it would be_), whereas the regression here would have a wider surface area of affected users and I wasn't sure how strict you'd be with regards to that.

I'll introduce allocations to avoid the mutation concern then.

epage: > > The mutation wasn't obvious even with `&mut`.
> 
> Could you please clarify?
> - Does `&mut filename` not indicate an intent for mutation to consider?
> - While I agree the actual mutation doesn't stand out in the function body due to no assignment, but a `set` method call, this was the same mutation as the prior code was using already prior, just the existing code was more implicit.

I saw a combinator and saw the function name and never looked at the parameter list because they should pure.


> I'll introduce allocations to avoid the mutation concern then.

I did a pass through this code today, applying some cleanups. However, I find the current general structure easier to understand than when inverting things. The `is_file` check creates very distinct requirements that drive the rest of the behavior and that should be highlighted rather than buried and smoothed over by conformity.

epage: I should add that I'm not going to debate this any further. What was proposed as a cleanup of a single function has taken too much focus proportionate to any theoretical gains.

polarathene: > The `is_file` check creates very distinct requirements that drive the rest of the behavior and that should be highlighted rather than buried and smoothed over by conformity.

So does `format_hint.is_some()` 🤷‍♂️ Both have a fallback on the other.

What is not ideal is the interleaving of the two format types, which is what you get when `is_file` is prioritized first.

> I did a pass through this code today, applying some cleanups. However, I find the current general structure easier to understand than when inverting things.

Strongly disagree, you have mixed use of `extensions()` and `file_extensions()`, the reader would need to familiarize with both and why these differences occur, and the implications of that should they be doing anything with this logic.

https://github.com/rust-cli/config-rs/blob/35bafb97f175fc1fba5e1244f7ea51d461b6b9bb/src/file/source/file.rs#L61-L82

You could still have had that consistent (_as [this commit demonstrates](https://github.com/rust-cli/config-rs/pull/684/commits/aa915d6f77ac48e59ab784528add947019c7eec1), wasted effort/time on my part clearly, despite your request for the staged commits_).

[Your commit](https://github.com/rust-cli/config-rs/commit/fd28355ca4f892e98c611aa7db2af05f6c827511) addressed what I was going to PR after this, thanks :)

---

> What was proposed as a cleanup of a single function has taken too much focus proportionate to any theoretical gains.

That's fully on you. I am certain I sunk far more time into it, only to have it rejected.

The bulk of all of this was a quibble over a concern with how I went about mutating `filename` that you chose to be incredibly stubborn over. Even after I reluctantly respect your request for changes, I return to find the PR rejected.

I do not appreciate how you handled this review.
- No positive engagement to the contributor experience.
- Requests to split commits for the benefit of a diff you can better follow (wasted time given the outcome)
- Blocking on a concern stubbornly, which once I return with changes to accommodate your request.... I find you've subsequently [push your own changes](https://github.com/rust-cli/config-rs/pull/688) and rejected the PR on the basis that it was too costly to time (_with no indication of this being an issue earlier, since you expressed you'd ignore the bulk of my responses when they're verbose_).

FWIW, "theoretical gains" may be applicable reasoning to this PR itself, but the decision and experience should account for influence of future contributions. I say this not just in regards to myself, but to consider with any other contributors.

I do understand your busy and must prioritize time. I do acknowledge the intent to be professional, but the communication / engagement has proven to be something I cannot trust on my own end to feel comfortable trying to engage more in the project as a contributor.

I don't want to keep head butting and debating. If I cannot trust you to communicate properly, and get my own time and efforts discarded... then there's no point in engaging in the project further for me. I don't want to invest more time into satisfying your review requests when it still risks outcomes like have occurred here.

polarathene:
Below can be ignored, this was being filled out prior to https://github.com/rust-cli/config-rs/pull/684#discussion_r2308830905 and the subsequent rejection of my efforts to contribute.

---

> I explicitly said otherwise, see [#684 (comment)](https://github.com/rust-cli/config-rs/pull/684#discussion_r2307736948)

Apologies, I wrote one comment after the other and had not been aware of a quick response from you as there was no real-time update on the page nor notification on my phone received until after posting the 2nd comment.

---

> this doesn't even get into the brittleness of relying on matching the behavior of the bool with the mutation.

I'm not sure what brittleness you're referring to?
- Nowhere else will the method be called it's not for public use
- A single closure is produced to be used in either `.filter()` or `.find()`
- The predicate short-circuits both in the `.filter()`/`.find()` iteration, and the internal `.any()` predicate.
- `filename` is mutated with the extension placeholder within the method producing the predicate closure, then captured into the returned closure for mutating with `.set_extension()`.
- When the mutated filename exists we return true.
- When nothing exists there is no use of `filename`, a `None` triggers the failure condition returning an error instead.

What is brittle there? As mentioned previously the burrow checker and related changes to control-flow ensure the scope of the mutation is localized and will raise compile failures should that change.

There is one concern where I think you have a point, should someone attempt to refactor this to remove the `match` statement, and shift the `None` match arm to a `format_hint.filter( ... ).or_else( ... )` such that instead of bailing on failure when `format_hint` returned `None` on the same predicate in `.filter()`, the mutation of `filename` will be a bug. I highlighted this concern in one of the staged commits (_when still transitioning from imperative code which required workarounds due to burrow checker_).

My response to that would simply be a test case that would catch it would flag it quite easily, hence with nothing else to deem it brittle, I disagree 🤷‍♂️ If you're aware of an actual problem I've overlooked I would love to know.

---

> This was not me rushing the review.

I don't know what to infer otherwise.
- You saw the `&mut` and dismissed the notion of a mutation occurring, originally saying the method name could have indicated it better 😕
- The function body itself was only a few lines, using roughly the same logic as existing code does, including the mutation `.set_extension()`.

If anything the scope of the mutation was far better communicated and visible than it was when everything was inlined in `find_file()`, so much so that we had explicit commentary to raise attention about this to the reader.

---

Regardless, I'll address the nitpick.

I hope you find no issue in an additional commit that switches `.filter()` => `.and_then()` and `.find()` => `.find_map()`? I'll take `&PathBuf` instead of `&mut PathBuf`, then create a new `PathBuf` for every format tried.

Below is the proposed `find_file()` (_after initial `filename` is set_), let me know your preferences and I'll commit it, unless the below still remains unacceptable 😭

```rust
// Fallback to checking compatibility with internally supported formats (`FileFormat` enum):
let mut internal_formats = all_extensions().keys().cloned();

// Ideally there is an exact filename match with a format hint, otherwise:
// - Without a hint => Try to identify the format via the file extension
// - Without an exact filename => Check if the filename exists with any known format extensions
if filename.is_file() {
    return match format_hint {
        Some(format) => Ok((filename, Box::new(format))),
        None => {
            let identify_format = |ext: &std::ffi::OsStr| {
                let ext = ext.to_str()?;
                internal_formats.find(|f| f.file_extensions().contains(&ext))
            };
            
            let format = filename
                .extension()
                .and_then(identify_format)
                .ok_or_else(|| self.error_invalid_format())?;
             
            Ok((filename, Box::new(format)))
        }
    }
}

// Preserve any extension-like text within the provided file stem by appending a fake extension
// which will be replaced by `set_extension()` calls (e.g. `example.file.placeholder` => `example.file.json`)
let mut filename = filename;
if filename.extension().is_some() {
    filename.as_mut_os_string().push(".placeholder");
};

match format_hint {
    Some(_) => format_hint.and_then(find_file_with_format(&filename)),
    None => internal_formats.find_map(find_file_with_format(&filename)),
}
.ok_or_else(|| self.error_invalid_path())
```

With the `file_exists_with_format()` method adjusted to (_**Note:** the `as_mut_os_string()` block could optionally be shifted into here after the clone_):

```rust
fn find_file_with_format<F: FileStoredFormat + Format + 'static>(
    filename: &PathBuf,
) -> impl Fn(F) -> Option<(PathBuf, Box::<dyn Format>)> {
    |format| {
        let mut filename = filename.clone();

        let file_exists = format.file_extensions().iter().any(|ext| {
            filename.set_extension(ext);
            filename.is_file()
        })

        file_exists.then_some((filename, Box::new(format)))
    }
}
```

That's still technically mutating `filename` within a predicate for `.any()`, which I assume would be rejected for the same reasons you're against this for my prior use via `.find()`/`.filter()`...

If so I can swap that out for the more verbose equivalent?:

```rust
for ext in format.file_extensions() {
    filename.set_extension(ext);

    if filename.is_file() {
        return Some((filename, Box::new(format)));
    }
}

None
```

Personally that doesn't read as well, but neither does the `find_file()` revision to accommodate a concern that is arguably minor given the context I've provided.

---

Here's the equivalent `find_file()` change if preferring the proposed change to avoid interleaving the two format types.
- This assumes `as_mut_os_string()` was moved into `find_file_with_format()`.
- The `else` branches no longer unify the return value with their `if` branch anymore, since they now return a tuple option instead.

```rust
match format_hint {
    Some(format) => {
        if filename.is_file() {
            Ok((filename, Box::new(format)))
        } else {
            find_file_with_format(&filename)(format)
                .ok_or_else(|| self.error_invalid_path())
        }
    }

    // Fallback to checking compatibility with internally supported formats (`FileFormat` enum):
    None => {
        let mut internal_formats = all_extensions().keys().cloned();

        if filename.is_file() {
            let format = internal_formats
                .find(identify_format(&filename))
                .ok_or_else(|| self.error_invalid_format())?;

            Ok((filename, Box::new(format)))
        } else {
            internal_formats
                .find_map(find_file_with_format(&filename))
                .ok_or_else(|| self.error_invalid_path())
        }
    }
}
```

src/file/source/file.rs
epage: > No positive engagement to the contributor experience.
>
> That's fully on you. I am certain I sunk far more time into it, only to have it rejected.

While I think its important for reviews to be a conversation, that the submitter should push back at times, you chose to persist well past that point and to put all of this time into arguing for the change.  This is clearly not a quibble when I was able to show actual harm from it.  Even if I hadn't, I still disagree with this being a quibble.  A lack of empathy was shown when I showed initial concern over the change and persisted even after I showed that harm all the way to this final post.

> Requests to split commits for the benefit of a diff you can better follow (wasted time given the outcome)

Cleaning up commits is not busywork but about taking responsibility for clearly communicating what is happening rather than foisting the burden on a maintainer for them having to reverse engineer what the change is and what motivates it.  Without this, I was not going to engage deeply enough with the change to be able to identify any ways for positive engagement.

I also gave this feedback in parallel to my feedback about pure combinators. I had assumed you would close out that feedback first as that had the potential to have a large impact on where this PR went but instead you persisted in pushing that change no matter how many times I said no and as I had to be more and more explicit about the fact of that part of the change.

> Blocking on a concern stubbornly, which once I return with changes to accommodate your request.... I find you've subsequently https://github.com/rust-cli/config-rs/pull/688 and rejected the PR on the basis that it was too costly to time (with no indication of this being an issue earlier, since you expressed you'd ignore the bulk of my responses when they're verbose).

#688 was born from me putting in extra of time to explore the problems with the code to see if I had alternative solutions to address your concern the underlying concern that led to this PR.  In doing that is when I noticed that I disagree with the direction fundamentally, something that was missed because the PR was not setup for clear communication.

polarathene: I don't expect you to read any of the below or respond to it, so no worries if you cannot afford the time 👍 

---

What's done is done, but if you're interested in the feedback below have at it. Hopefully it provides some perspective from a contributor standpoint, that it would improve how you engage with future contributors.

Neither of us feel in the wrong with how we participated in the PR, that's unlikely to change. I don't expect you to express any empathy for how you treated my contribution and time, but I do acknowledge my communication could improve, I sought to provide you with as much context as I would appreciate to receive, whilst already getting a read early-on that my verbosity/engagement would not be well-received and misinterpreted.

---

On reflection, I need to invest more time into noticing this and progressively disclosing such details to avoid overwhelming the reader and running into miscommunication.
- It would be more optimal if I engaged tersely and steered conversation with brief questions to garner when certain information would actually be appreciated.
- I mean no offense when I make this assumption, but usually this conflict in communication style is with allistic individuals which requires great care to avoid miscommunication.

So while I am disappointed with the outcome (_not so much the rejection of this specific PR, but the experience as detailed below_), and my decision to cease returning to pursuing contributions to `config-rs` (_as throwing away all the time I invested in my stalled PRs was not easy to let go_), I have found comfort in relieving myself of that burden of commitment that's nagged at me over the past two years (_I still have other projects with stalled commitments to resolve over the past couple years_).

I love contributing back to the community with OSS, but as you may be aware it can take a toll, I've yet to even enter the industry professionally as I don't feel I have the expertise yet to be worth paying for my time, my communication skills clearly need to improve as this isn't the only interaction I've had contributing where I've failed at collaboration (_one RedHat backed project blocked me from their entire organization on GH because I tried to improve the clarity on their documentation with reproducible proof and references to justify it, but misunderstanding and prejudice from the reviewer had them ban my account_). It's disheartening when it's that bad, but thankfully it's rare to encounter in the community 😅 Still I am learning how to better handle those situations or when it's more prudent to distance myself to avoid escalating any risk of tensions further.

---

>  This is clearly not a quibble when I was able to show actual harm from it

You did not at any point demonstrate harm from my use of `&mut`, only that you did not initially notice a mutation and were somehow surprised despite stating you had clearly observed the presence of `&mut filename` and intentionally chose to think nothing of it 🤷‍♂️

Prior to revealing that you had observed `&mut`, you had complained about the method name not conveying mutation. I noted the `&mut` visibility clearly conveys intent to mutate and queried how addressing the method name concern would be any improvement, to which you conveniently ignored.

You were fixated on a concern with no tangible harm in it's usage. The only problem with it actually noted, was cited by me and that a test case would very easily resolve that.

---

> A lack of empathy was shown when I showed initial concern over the change and persisted even after I showed that harm all the way to this final post.

To accuse me of a lack of empathy when I've demonstrated empathy through my responses (_which you've openly admitted to choosing to gloss over / ignore_) is ridiculous when your attitude towards me through the post has been pessimistic with no appreciative feedback what so ever with my contribution.

You had nothing positive to say with any of the changes, instead fixating on the mutation quibble with a disregard to any valid points I raised on my end to support the choice. You actively engage in the debate over the concern, only to point the finger at me as wasting your time discussing the nitpick and closing the PR.

There was no warning or communication that such an action would take place if I engage in discussion openly. I even [agreed to address the concern](https://github.com/rust-cli/config-rs/pull/684#discussion_r2308800175) you were so stubborn over, since I'd rather push the PR forward, yet despite communicating this to you, you go ahead and shut the PR down.

That was **not collaborative**, that was **not a positive experience with empathy** on your end. That was **disrespect** and leveraging your position of authority.

Had I not been willing to accommodate your concern, I would fully agree with the right to block and shut down my contribution.

---

> you chose to persist well past that point and to put all of this time into arguing for the change.

> you persisted in pushing that change no matter how many times I said no

This is a common miscommunication that I encounter with allistic / neurotypical people, who have an entirely different perspective of how I'm engaging with them. Where I try to ensure that we are on the same page by asking questions and providing further context, it tends to be perceived as arguing/challenging, or making excuses.
- Attempting to even express this difference or feedback related to such observations with such individuals for some reason often leads to dismissive/deflective behaviour or increased tension 🤷‍♂️
- Likewise I may come across as insisting that I am right and you are wrong. I don't mind being mistaken and accepting that, what I care about is being correct. For things to make logical sense. So I seek further context if I'm lacking it, or I try to provide further context to the other party if they appear to have trouble understanding my statements when their own statements have gaps that they for whatever reason refusing to elaborate on when I request context on their end. The goal is to align on what is correct, or respectfully disagree when evidence is lacking to resolve (_leaving us to resort to biases_).

The discussion over the mutation concern was explaining my reasoning, and trying to better grasp your own concern and how you encountered a scenario where mutation was not obvious to you (_but somehow was with the code prior to the PR where it was more implicit_).

I could accept letting it go, but I was adamant about expressing the advantages while demonstrating the disadvantages without it, and how it could work safely so both parties are happy. You just chose to remain close minded over it and conveniently ignored any valid points I made.

> Cleaning up commits is not busywork but about taking responsibility for clearly communicating what is happening rather than foisting the burden on a maintainer for them having to reverse engineer what the change is and what motivates it.

I had no issue in splitting out commits and the extra work that involved to accommodate you. I do understand the value of such as a maintainer that reviews PRs too, but in the context of this PR it made little sense and I was left with the impression that you didn't even bother to review commit by commit after that effort was made.

From my perspective I made the effort and it went no where, had no impact. You remain fixated and stubborn on a point that I realized was going no where.

---

Even after splitting the commits for you and requesting how you'd like me to approach an alternative, and citing concerns related to how I'd approach that you ignored that and proceeded with your own changes and subsequent closing of this PR, just as I [returned with a few snippets of how I thought I could best accommodate your need to avoid mutation](https://github.com/rust-cli/config-rs/pull/684#issuecomment-3235651823) after confirming that introducing extra allocations would be acceptable.

Both of those efforts, and the entire contribution effort prior to that felt completely and utterly wasted. I explored many revisions prior to submitting this PR, I did not anticipate you'd fixate on the mutation given it was already happening prior to the PR and my responses providing justification to support it.

**You were vague in review feedback:**
- > The name for this function does not communicate how it mutates `filename`
- > I missed the mutation.
- > I did not expand because the point was that it failed your assumption that the intention was obvious or easier to grok. 
- > The current code structure will not be merged.
- > > _You did not start from the call-site with a clear indication of mutation via an explicit `&mut` passed in?_
  > I did. The mutation wasn't obvious even with `&mut`.
- > I saw a combinator and saw the function name and never looked at the parameter list because they should pure.
- > I feel combining filtering with mutation to be brittle and non-obvious. In general, I feel that iterator combinators should not do mutations.

At no point was an alternative suggested or an explanation for why mutation prior to the PR was deemed acceptable. The most I could identify from the feedback was that `.filter()` + `.find()` usage with the predicate wasn't even the issue, it was the mutation within `.any()` where your same reasoning would have applied, but you didn't seem to care about that specifically (_where the function name and predicate closure had no relevance as this concern occurred within the small closure itself_).

I raise this point [here](https://github.com/rust-cli/config-rs/pull/684#issuecomment-3235651823) where I state that the proposed change with allocations that avoids mutation where you had cited it as an issue was probably still not acceptable, where I compare it to the imperative `for` loop equivalent of `any()` combinator. Something you're completely okay with, presumably due to semantic differences.

So all I was left with was "I don't like this because it _feels_ brittle", doesn't matter how much I harp on about Rust avoiding side-effect concerns far better and how we can reliably avoid your concern.

Along with a broader statement of "current code structure", which **dismissed the entire PR as not being a worthwhile revision** due to one much smaller scope of change that you refused to budge on or openly discuss how to better approach it, even after I split the commits and verbosely broke down the `find_file()` branches and concerns I was trying to address, all so you could better grok the functionality for review and **collaborate on a way forward that works for you**.

---

> In doing that is when I noticed that I disagree with the direction fundamentally, something that was missed because the PR was not setup for clear communication.

The [opening PR message](https://github.com/rust-cli/config-rs/pull/684#issue-3358450271) detailed changes and flow fairly well, and a even simpler overview.

That was not sufficient for you, you needed the commit split up so you could better follow along with how I detangled the concerns I raised in that description. I did that for you, and provided a more verbose walkthrough in feedback comments.

I honestly don't know how much of that you bothered to digest, and I don't blame you but I tried very hard from the start to try make this easy for you as the reviewer to understand the functionality and control flow given that you'd rather not waste time getting familiar with it manually.

The PR itself if viewed solo has rather clear control flow, the concerns I raised and addressed are clearly handled there. Many readers should be able to look at that (_and without getting fixated over this mutation concern for a moment_) understand what the logic is doing and rather quickly grok how all 4 variations are processed.

This was not the case with the existing code prior to the PR. You'd get a rough idea from the disjointed flow, but some logic was lacking context and if you tried to work on it without familiarity, you'd run into several points of friction from assumptions. You've at least gone through a similar process yourself and cleaned that up a bit since, but some concerns still remain.

Anyway, you can state that I communicated poorly.
- Whenever you expressed a need for more information or presenting it differently, I did my best to accommodate and clearly walk you through any changes/motivations.
- Even the commit messages had some detailed context for intermediary diffs (_which personally those commits seemed redundant, but I don't know how much granularity you expected for a small function removing interleaving of types in control flow_).

If you were to have communicated better about what you didn't understand (_since as the quote notes, you lacked the perspective you needed until getting hands-on_), I would have tried to assist. You were able to nitpick on a concern that was pragmatically negligible, whilst lacking the ability to suggest what the alternative would look like for you leaving me to play guessing games or seek more context.

PR #685, opened 2025-08-27
fix: Run tests without depending on `openssl`

polarathene:
`reqwest` is only used for a crate example basic HTTP call:

https://github.com/rust-cli/config-rs/blob/1483f4cfc0169c857d5a7d551c7fab02eeb17f06/examples/async_source/main.rs#L61

When cloning the project repo and running tests, it'd fail without OpenSSL devel package present on the system to build `reqwest`. I was going to replace that with `rusttls` but it seems HTTPS isn't required at all so opt-out of default features seems more prudent?

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -99,12 +99,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "atomic-waker"
-version = "1.1.2"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "1505bd5d3d116872e7271a6d4e16d81d0c8570876c8de68093a09ac269d8aac0"
-
-[[package]]
 name = "autocfg"
 version = "1.1.0"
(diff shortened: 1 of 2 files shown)


epage:
Fine with this change. CI needs addressing

polarathene:
> CI needs addressing

Is there a preferred way to update `Cargo.lock`? Or just `cargo update` is ok for this project? Does that need to be run from Rust 1.75, or can it be done from a newer toolchain?

Some of the other `Cargo.toml` deps could probably be updated given the Rust 1.75 MSRV which would remove some duplicate crates, `warp@0.3` for example could go to `warp@0.4`. It's also a dev-dependency but I'm not sure if you'd want it bundled in this PR or not? (_if so do you want it all in a single commit?_)

epage:
Run `cargo check` and ~~commit~~ amend your commit.

Issue #621, opened 2024-12-19
Re-evaluate default features


epage:
`async` should not be a default feature.

`convert-case` needs evaluation.

As for file formats, I don't think `ini`, `ron`, `yaml`, and `json5` should be defaults but there is the question of what formats should be.

polarathene:
Might be good to reference what alternative crates like [Figment](https://github.com/SergioBenitez/Figment/blob/12dd1bc96bdd69417803bc794b6953d89c3a1af1/Cargo.toml#L17-L18) and [Confique](https://github.com/LukasKalbertodt/confique/blob/5b240713241434d455d4d444575dfee2ca6d1c63/Cargo.toml#L37-L38) are doing? (_both links to their `Cargo.toml` default features, practically empty_)

Makes sense to just bring in the crate and then opt-in to the file format support you need? I'm not too familiar with the cost of having the different formats by default, I assume DCE works to reduce bloat (_might be reliant on LTO?_), but prior to that any deps are still built at compile-time which redundantly uses resources like time, disk, bandwidth. Is that the main concern?

epage:
The fact that we [dynamically resolve file stems to file names](https://docs.rs/config/latest/src/config/file/mod.rs.html#127-130) would be a major inhibition against dead code elimination.  Even if that feature wasn't used, it would likely take a Sufficiently Smart Compiler to propagate just the right constants through to eliminate unused file formats.

Costs for extra file formats
- Build times and binary size
- We implicitly extend applications compatibility surface

PR #694, opened 2025-09-23
feat: add corn file format

JakeStanger:
This adds `corn` as a new file format.

I've enabled it by default as per the current standards, but did see discussion in #621 as to possibly changing this approach.

https://cornlang.dev

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -222,6 +222,7 @@ dependencies = [
  "glob",
  "indexmap",
  "json5",
+ "libcorn",
  "log",
  "notify",
  "pathdiff",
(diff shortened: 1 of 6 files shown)


epage:
Would prefer for us to stop adding to `default` now and we'll remove from it in the future.

JakeStanger:
Cool, I've dropped it from the default features list

JakeStanger:
Updated the commit message to comply. Looks like the main CI job failure is unrelated?

epage:
Unsure how CI got into a red state but I just fixed it and rebased your branch on top of those fixes

JakeStanger:
Cheers for merging! I've just realised I didn't touch the readme. Want me to update that in or are you happy to?

epage:
I just pushed it

Issue #532, opened 2024-02-08
ux: Lack of key and origin in ConfigError::Message leads to uncertain sources from serde error messages


nmathewson:
## Summary

If a configuration error passes through a `serde::de::Error` implementation, it is wrapped as a `ConfigError::Message`, and information about what caused the error is not preserved.  This confuses the users, because the message does not tell them which option was set incorrectly.

## Example

```
use serde::Deserialize;

#[derive(Deserialize, Clone, Debug)]
pub struct Settings {
    direct: Option<u8>,
    via_serde: Option<Item>,
}

#[derive(Deserialize, Clone, Debug)]
#[serde(try_from="String")]
pub struct Item(u8);

impl TryFrom<String> for Item {
    type Error = std::num::ParseIntError;
    fn try_from(s: String) -> Result<Item, Self::Error> {
	s.parse().map(Item)
    }
}

fn main() {
    let settings = config::Config::builder()
	.add_source(config::File::with_name("demo.toml"))
	.build()
	.unwrap();

    let settings = settings
	.try_deserialize::<Settings>();

    println!("{:?}", settings);
}
```

If the above code is run with the configuration `direct = "foo"`, then we get a helpful error message:
```invalid type: string "foo", expected an integer for key `direct` in demo.toml```

But if the above code is run with the configuration `via_serde = "foo"`, then we get the not-so-helpful message:
```invalid digit found in string```.
Note that it doesn't mention "via_serde" or "demo.toml".

It would be great if instead we go something like:
```invalid digit found in string (for key `via_serde` in demo.toml```

## Possible solutions

My first thought here would be to adjust Message to include an optional `key` fields, so that `prepend_key` can set it. 

If we don't want to break backward compatibility with code that uses the current Message variant, I would define a new MessageWithKey or InvalidValue error variant, and have prepend_key() convert Message into that variant instead.

## Reference

For us this issue is https://gitlab.torproject.org/tpo/core/arti/-/issues/1267

nmathewson:
If you like, I am happy to work on any of the solutions here—just let me know which approach you would prefer.

epage:
#695 originally covered several cases but pulled back due to some of them requiring a breaking change.

PR #695, opened 2025-09-23
feat: Improve errors from try_deserialize

barskern:
Enables user code to handle missing and invalid fields after calling try_deserialize instead of having to match and search in the string returned in `ConfigError::Message`.

Personally I deem the `String::leak` call safe for my usage, as I would expect to terminate my program on configuration errors anyhow, however we could certainly also change the `expected` field to be a `Cow<str>` or similar to handle both `&'static str` and `String` elegantly.

Part of #532

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -230,7 +230,7 @@ impl fmt::Display for ConfigError {
             ConfigError::Foreign(ref cause) => write!(f, "{cause}"),
 
             ConfigError::NotFound(ref key) => {
-                write!(f, "configuration property {key:?} not found")
+                write!(f, "missing configuration field {key:?}")
             }
 
             ConfigError::Type {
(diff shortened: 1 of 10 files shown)


src/error.rs
```diff
@@ -289,6 +317,33 @@ impl de::Error for ConfigError {
     fn custom<T: fmt::Display>(msg: T) -> Self {
         Self::Message(msg.to_string())
     }
+
+    fn missing_field(field: &'static str) -> Self {
+        Self::NotFound(field.into())
+    }
```
epage, line 295: I'm willing to accept this now; the others will need to wait until a breaking change.

I would also prefer us to update the error message for this to better match serde's text.  I feel our own message is confusing.  Please change the message in a separate commit.

barskern: Sure, I've made some changes. I wasn't really sure how to change the message, but I tried to keep it inline with serde while still keeping it relevant for the configuration aspect of it.

epage:
Looks like some more test cases need updating

barskern:
> Looks like some more test cases need updating

Yeah, I didn't run with `--all-features`. While going through it I also changed some of the places were we can now rather match on the enum than search in the string, to show that it should return a proper variant.

epage:
Thanks!

PR #703, opened 2025-10-08
feat(env): underscore field nesting for single `_` separators

aleris:
Allow to nest with single `_` even when field names contain underscores via `.underscore_nesting(true)``.
Can be used when field names also have `_` with environment overrides.
Can be used as an alternative for double underscore separator '__' for nesting.
Supports deep nesting. Is off by default.


src/env.rs
```diff
@@ -154,8 +172,24 @@ impl Environment {
...
+    /// # use serde::Deserialize;
+    /// #[derive(Clone, Debug, Deserialize)]
+    ///   struct MyConfig {
+    ///     pub my_string: String, // will be parsed as String
+    ///     pub my_list: Vec<String>, // will be parsed as Vec<String>
+    ///   }
+    /// let source = Environment::default()
+    ///     .list_separator(",")
+    ///     .with_list_parse_key("my_list");
+    /// ```
```
aleris, line 192: unrelated but to clarify documentation which was  a bit confusing for me

tests/testsuite/env.rs
```diff
@@ -518,6 +518,117 @@ fn test_parse_string_and_list_ignore_list_parse_key_case() {
     );
 }
 
+#[test]
+fn test_parse_nested_double_separator() {
```
aleris, line 522: without kebab case conversion, this is to ensure it works in simple mode as a mirror of the next test which enables single underscore nesting

epage:
Note that our contributing guide asks to start with Issues, rather than PRs.  Moving this to a Draft until there is a mutual understanding and agreement on the problem and the solution.

aleris:
Ah, ok, thanks, will follow up. 
Regarding starting with issues, maybe would help to change the language to be more explicit instead of current "it is probably best" and "It might be good" which suggest is rather optional (or I least this is how I read it).
https://github.com/rust-cli/config-rs/blob/main/CONTRIBUTING.md#pull-requests

aleris:
added https://github.com/rust-cli/config-rs/issues/704

PR #706, opened 2025-10-10
Add support for environment substitution within config files

awgroeneveld:
Coming from a Spring background, I'm missing the possibility of expansion of environment variables referenced from within configuration files, giving an even more flexible approach next to overriding properties which is in itself a great functionality of the config crate. Using this proposed change, there is no requirement to fully-qualify environment variable names that override the config files.

Spring also has support for a default value.

Both elements have been implemented in this PR, where :- splits the environment variable from the default, e.g.:
```yaml
database:
   host: ${DBHOST:-127.0.0.1}
   port: ${DBPORT:-27017}
```
Due to current design, this had to be implemented in the file format files.
It's currently implemented as a feature in order not to break existing code. The feature is substitute_env

Although there are some crates that could be used like shellexpand and envsubst this was the easiest way to prevent more dependencies.

The implementation is done by creating an extension function by means of trait ExpandEnvVars which has a default behavior if the feature is not activated.
The trait impl has its own tests. Furthermore, base upon the existing yaml and toml tests soms simple tests have been provided in [file_with_env_vars.rs.]

Would you mind considering the proposal? I'd be glad to head from you.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -121,7 +121,7 @@ pre-release-replacements = [
 ]
 
 [features]
-default = ["toml", "json", "yaml", "ini", "ron", "json5", "convert-case", "async","corn","substitute_env"]
+default = ["toml", "json", "yaml", "ini", "ron", "json5", "convert-case", "async"]
 json = ["serde_json"]
 yaml = ["yaml-rust2"]
 ini = ["rust-ini"]


epage:
> Would you mind considering the proposal? I'd be glad to head from you.

In [CONTRIBUTING.md](https://github.com/rust-cli/config-rs/blob/main/CONTRIBUTING.md#pull-requests), we ask for proposals to come in the form of Issues.  They should focus on your use case and the problem you are having within it.

PR #708, opened 2025-10-15
Allow creating `File` with `PathBuf`

TheTollingBell:


epage:
To avoid making bad guesses, could you explain the motivation in your PR description?

PR #709, opened 2025-10-20
perf: pre-allocate arrays and maps

nik-rev:


epage:
How much of a difference does this make in practice?

nik-rev:
For 1 million TOML keys, it took `8.82` seconds and now it takes `8.25` seconds to call `Config::build()`. 6.5% increase.

Feel free to close the PR if you think it's not worth it. I needed to understand a region of code in this crate and found a possible, albeit minor, improvement

# benchmark

```rust
#[test]
#[cfg(feature = "toml")]
fn benchmark() {
    use std::time::Instant;
    use config::{Config, File, FileFormat};

    let config = (0..1_000_000)
        .map(|i| format!("key{i} = {i}\n"))
        .collect::<String>();

    let c = Config::builder().add_source(File::from_str(&config, FileFormat::Toml));
    let before = Instant::now();
    let c = c.build();
    println!("{:?}", before.elapsed());
}
```

Outputs (with this PR):

```
8.200018192s
8.270070666s
8.27713242s
```

Before this PR:

```
8.779736576s
8.760608855s
8.913664742s
```

epage:
I think I'll hold off.  If we want to improve this, likely the best way is to deserialize directly to our own type

Issue #16, opened 2017-02-11
Support dotenv file format


dtolnay:

mehcode:
Didn't think someone would actually want this when I thought through it more. The dotenv lib could be used alongside this one easy enough.

dtolnay:
Cool it's up to you. I just wanted to make sure this is tracked somewhere because you listed it in the [roadmap](https://www.reddit.com/r/rust/comments/5pzjwo/configrs_layered_configuration_system_for_rust/dcv2nde/) but there was no ticket.

mehcode:
I'm still planning to do this. There are many tools out there that use "generic" `.conf` files and this would satisfy that. I don't want to call the format `Dotenv` though. Perhaps `Shell`.

mehcode:
And I just found a good use case for supporting specifically the `.env` file. Diesel will only support loading the database URL from the environment until `#[proc_macro]` stabilizes. For now I can keep just the `DATABASE_URL` in `.env` and load it into my configuration stack via:

```rust
c.merge(config::File::new(".env", config::FileFormat::Shell)).unwrap()
```

Remember that we don't support the above format yet. If anyone wants to work on it, follow this spec: https://github.com/bkeepers/dotenv#usage and use nom. Probably should be its own crate and with coordination from @slapresta, perhaps it can replace the hand-rolled parser in rust-dotenv (lots of TODOs in the parser so I'd guess @slapresta would be open to it).

slapresta:
Hi!

Improvements to the parser (which is basically a regex plus quote handling logic) are definitely welcome. Reading from a specific file should be feasible with dotenv's`from_file`; the problem is that dotenv automatically adds those to the environment. It should be feasible to split [the part that adds this to the environment](https://github.com/slapresta/rust-dotenv/blob/master/src/lib.rs#L195-L199) to a separate function, and use the `ParsedLine`s by themselves to construct your config structures.

azzamsa:
> Didn't think someone would actually want this when I thought through it more. The dotenv lib could be used alongside this one easy enough.

Any update about this decision?
Even though config-rs refuse to load `.env` to system environment. At least we should support *reading from .env file* (Currently we support JSON, TOML, YAML, HJSON, INI). So that we don't need `dontenv` alongside config-rs.

porfirioribeiro:
I needed this for myself so i created a little adapter to loading dotenv files
Might be useful for someone, so I'm dropping it here:
```rust

[...]

PR #710, opened 2025-10-22
feat(dotenv): Adding support for dotenv files

chickencoding123:
This adds support for [dotenv](https://www.dotenv.org/docs) file format. It is accomplished by using [dotenvy](https://crates.io/crates/dotenvy) crate. There's an open issue asking for this feature here https://github.com/rust-cli/config-rs/issues/16.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -148,6 +148,7 @@ convert_case = { version = "0.6.0", optional = true }
 pathdiff = "0.2.3"
 winnow = "0.7.15"
 serde-untagged = { version = "0.1.9", optional = true }
+dotenvy = { version = "0", optional = true }
 
 [dev-dependencies]
 serde = { version = "1.0.228", features = ["derive"] }
--- a/src/file/format/dotenv.rs
+++ b/src/file/format/dotenv.rs
@@ -4,6 +4,9 @@ use std::io::Cursor;
 use crate::map::Map;
 use crate::value::{Value, ValueKind};
 
+#[cfg(feature = "convert-case")]
+use convert_case::{Case, Casing};
+
 pub(crate) fn parse(
     uri: Option<&String>,
     text: &str,
@@ -11,7 +14,13 @@ pub(crate) fn parse(
     let mut map: Map<String, Value> = Map::new();
     let str_iter = dotenvy::Iter::new(Cursor::new(text));
     for item in str_iter {
-        let (key, value) = item?;
+        let (mut key, value) = item?;
+
+        #[cfg(feature = "convert-case")]
+        {
+            key = key.to_case(Case::Snake);
+        }
+
         map.insert(key, Value::new(uri, ValueKind::String(value)));
     }
 
--- a/tests/testsuite/file_dotenv.rs
+++ b/tests/testsuite/file_dotenv.rs
@@ -16,8 +16,8 @@ BAZ=qux
         .build()
         .unwrap();
 
-    assert_eq!(s.get::<String>("FOO").unwrap(), "bar");
-    assert_eq!(s.get::<String>("BAZ").unwrap(), "qux");
+    assert_eq!(s.get::<String>("foo").unwrap(), "bar");
+    assert_eq!(s.get::<String>("baz").unwrap(), "qux");
 }
 
 #[test]
@@ -34,7 +34,7 @@ BAR=${UNDEFINED:-}
         .build()
         .unwrap();
 
-    assert_eq!(s.get::<String>("BAR").unwrap(), "");
+    assert_eq!(s.get::<String>("bar").unwrap(), "");
 }
 
 #[test]
@@ -55,8 +55,8 @@ BAZ=qux
         .build()
         .unwrap();
 
-    assert_eq!(s.get::<String>("FOO").unwrap(), "bar");
-    assert_eq!(s.get::<String>("BAZ").unwrap(), "qux");
+    assert_eq!(s.get::<String>("foo").unwrap(), "bar");
+    assert_eq!(s.get::<String>("baz").unwrap(), "qux");
 }
 
 #[test]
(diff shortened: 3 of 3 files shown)


Cargo.toml
```diff
@@ -121,12 +121,13 @@ pre-release-replacements = [
 ]
 
 [features]
-default = ["toml", "json", "yaml", "ini", "ron", "json5", "convert-case", "async"]
+default = ["toml", "json", "yaml", "ini", "ron", "json5", "dotenv", "convert-case", "async"]
```
epage: We're no longer accepting new `default` formats and in the next breaking change will reduce what we do have (see #621).

Cargo.toml
```diff
@@ -121,12 +121,13 @@ pre-release-replacements = [
...
 json = ["serde_json"]
 yaml = ["yaml-rust2"]
 ini = ["rust-ini"]
 json5 = ["json5_rs", "dep:serde-untagged"]
 corn = ["dep:corn"]
+dotenv = ["dotenvy"]
```
epage:
```suggestion
dotenv = ["dep:dotenvy"]
```
Otherwise, we'll expose a `dotenvy` feature flag

Cargo.toml
```diff
@@ -143,6 +144,7 @@ rust-ini = { version = "0.21.3", optional = true }
 ron = { version = "0.8.1", optional = true }
 json5_rs = { version = "0.4.1", optional = true, package = "json5" }
 corn = { version = "0.10.0", optional = true, package = "libcorn" }
+dotenvy ={ version = "0.15.7", optional = true }
```
epage:
```suggestion
dotenvy = { version = "0.15.7", optional = true }
```
(look forward to when `cargo fmt` can reformat `Cargo.toml`)

src/file/format/mod.rs
```diff
@@ -57,6 +60,10 @@ pub enum FileFormat {
     /// Corn (parsed with `libcorn`)
     #[cfg(feature = "corn")]
     Corn,
+
+    /// Dotenv (parsed with `dotenvy`)
+    #[cfg(feature = "dotenv")]
+    Dotenv,
```
epage, line 66: Does this work as another `FileFormat` / `Source`?  I suspect there is a lot of design overlap with [`Environment`](https://docs.rs/config/latest/config/struct.Environment.html) that we'd want to make this functionality on that.

chickencoding123: This is another `FileFormat`. You can see the usage from the tests (added snippet for ref. here). It is similar to existing file formats. Also `dotenvy` dependency will take care of the overlap, if any. However I ensured OS env take precedence as per existing file formats.
```rs
 let s = Config::builder()
        .add_source(config::File::from_str(
            r#"
                FOO=bar
                BAZ=qux
            "#,
            config::FileFormat::Dotenv,
        ))
        .build()
        .unwrap();
```

epage:
Feel free to modify your commits for how they should be merged

src/file/format/mod.rs
epage: FYI resolving conversations that aren't ready for it can make PR reviews more difficult on maintainers.  You should be pretty confident that you understood the concern and responded to it as the maintainer expected to resolve it.  That was not the case here.

Some traits of dotenv are
- Stringly typed (so is ini)
- Unstructured

Both of these characteristics are shared with `Environment` which has a lot of functionality for solving these problems.  In fact, this is another form of environment variables!  The only difference is it also shares some of the file loading characteristics of file formats.

src/file/format/dotenv.rs
```diff
@@ -0,0 +1,31 @@
...
+    uri: Option<&String>,
+    text: &str,
+) -> Result<Map<String, Value>, Box<dyn Error + Send + Sync>> {
+    let mut map: Map<String, Value> = Map::new();
+    let cursor = Cursor::new(text);
+
+    for item in from_read_iter(cursor) {
+        let (key, mut value) = item?;
+
+        let os_env_vars = std::env::vars_os();
```
epage: This should not be doing manual layering, that is what the `ConfigBuilder` is for.

chickencoding123: Dotenvy overrides OS var which is inconsistent with existing `FileFormat` in this crate. Manual layering is appealing here because file formats sit below `ConfigBuilder` in the hierarchy. Coupling a `ConfigBuilder` inside a file format will lead to difficult refactoring in the future.

I think for the purposes of this specific file format, simple env override should suffice?

src/file/format/mod.rs
epage: I've raised this (and other questions) at https://github.com/rust-cli/config-rs/issues/16#issuecomment-3437230286 as I prefer to work out high level designs in Issues rather than PRs.

src/file/format/dotenv.rs
epage: > Manual layering is appealing here because file formats sit below ConfigBuilder in the hierarchy. Coupling a ConfigBuilder inside a file format will lead to difficult refactoring in the future.

`Environment` and each `File` are manually added as sources and the user has control over the precedence order.  However we support DotEnv, it will be similar, so we don't need to do this here.  If we did,. it would belong in `Environment` which is what is handling the OS vars.

src/file/format/mod.rs
chickencoding123: I've updated to strike a balance between dotenv ambitions and preferences of this repository.

epage: I'm not aware of any dotenv "ambitions" that come in conflict with "preferences of this repository".  Also, for any "preferences" I'm giving, they are about fully supporting dotenv rather than doing it party way.

chickencoding123: 1. I had to dig into dotenvy source code to find out where they do not assume setting os::env. I found it ambitious to assume user always wants to set os::env
2. Preference for `config-rs`, this repository, is to use something like `.add_source(config::Environment::with_prefix("some prefix").separator("some separator"))` for handling env.

My use-case has been resolved already. Let me know about the next steps here.

epage: > Preference for config-rs, this repository, is to use something like .add_source(config::Environment::with_prefix("some prefix").separator("some separator")) for handling env.

We have set an expectation with users on what we support for environment variables.  Just because we now support reading them from a different source should not change those expectations.  This isn't a matter of "my taste is for doing something different".

> Let me know about the next steps here.

I feel I have set expectations.  If there is a question on them, feel free to ask questions.

Issue #717, opened 2025-11-18
json5 is abandoned


gtema:
json5 crate has been declared abandoned and there is https://rustsec.org/advisories/RUSTSEC-2025-0120

epage:
The other options do not seem very mature at this time so I think for now we will stick with `json5` as long as there isn't an identifiable problem with it.

polarathene:
FWIW the maintainer has chimed in that they're back to maintaining the project: https://github.com/callum-oakley/json5-rs/issues/51#issuecomment-3591976216

wdonne:
Version 1.3.0 is available.

PR #718, opened 2025-11-18
Replace json5 with serde_json5

liamadamson:
json5 is no longer maintained


liamadamson:
Addresses #717.

epage:
See https://github.com/rust-cli/config-rs/issues/717#issuecomment-3547960437

Issue #693, opened 2025-09-20
UpperCamelCase fails to capitalize first word immediately after a multilevel separator in Environment.


dtessman:
```
let mut builder = Config::builder()
.add_source(File::with_name(&config_path).format(FileFormat::Json).required(false))              .add_source(Environment::default().prefix("CONFIG").prefix_separator("_").separator("__").convert_case(Case::UpperCamel));
```

If the file contains:

```
Otel {
            Endpoint: "FROM FILE"
}
```

And the environment variable "CONFIG_OTEL__ENDPOINT" equals "FROM ENV"

getting "Otel.Endpoint" will equal "FROM FILE" and "Otel.endpoint" will equal "FROM ENV" rather than the environment overriding the file.

PR #746, opened 2026-03-19
Refactoring env var case convert to convert each segment between separator

11alieberman:
Currently, it looks like when you are parsing the environment variables into the config and you have some key with multiple segments separated by the ., (key1.key2.key3) when you do the case convert for a case like Pascal, it will result in (Key1.key2.key3). I am not sure if this is intended, because I would think the result should be (Key1.Key2.Key3), where each segment is individually case converted.

This PR aims to refactor the case conversion to convert each segment and not just the first segment.

Also aims to address https://github.com/rust-cli/config-rs/issues/693

### DIFF
--- a/src/env.rs
+++ b/src/env.rs
@@ -296,7 +296,11 @@ impl Source for Environment {
 
             #[cfg(feature = "convert-case")]
             if let Some(convert_case) = convert_case {
-                key = key.to_case(*convert_case);
+                key = key
+                    .split('.')
+                    .map(|segment| segment.to_case(*convert_case))
+                    .collect::<Vec<_>>()
+                    .join(".");
             }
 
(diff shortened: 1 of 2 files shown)


11alieberman:
@epage Would you have time to take a look at this?

epage:
This claims to be a refactor but claims to address #693 which would mean there is user-facing changes and so this wouldn't be a refactor anymore.

The linked issue's status is `S-triage`.  In general, we don't look at PRs and discourage the submission of them unless there is agreement on the problem and the proposed solution, see https://github.com/rust-cli/config-rs/blob/main/CONTRIBUTING.md#pull-requests

As such, I'm leaning towards closing this, not as a statement about what this PR does but that this is not the appropriate stage to be looking at code.

PR #755, opened 2026-05-13
Update json5 crate dependency from 0.4.1 to 1.3.1

notpeter:
This updates the JSON5 dependency from v0.4.1 (2021-09-21) to v1.3.1 (2026-02-27).

If you look at the diff of [0.4.1...1.3.1](https://github.com/callum-oakley/json5-rs/compare/0.4.1...1.3.1) you can see v1.0.0 was a clean rewrite which switched licenses (ISC to MIT) and switched from a PEST.rs generated backend to a hand-written one.

This results in a smaller download (net -1.7MB when dependencies included) and a slightly faster build (1-3% when building config-rs with default features on my M5 Pro Macbook).

According to the author, performance is ~20x better:
> `google/serde_json5` was a fork of the Pest version of `serde_json5`
> 
> serde_json5 benchmarks

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -832,13 +832,12 @@ dependencies = [
 
 [[package]]
 name = "json5"
-version = "0.4.1"
+version = "1.3.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "96b0db21af676c1ce64250b5f40f3ce2cf27e4e47cb91ed91eb6fe9350b430c1"
+checksum = "733a844dbd6fef128e98cb4487b887cb55454d92cd9994b1bafe004fabbe670c"
 dependencies = [
- "pest",
- "pest_derive",
  "serde",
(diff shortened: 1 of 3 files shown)


epage:
See
```
  error[E0658]: `let` expressions in this position are unstable
     --> /home/runner/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/json5-1.3.1/src/de.rs:271:15
      |
  271 |         while let Some((o, c)) = self.peek()
      |               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
      |
      = note: see issue #53667 <https://github.com/rust-lang/rust/issues/53667> for more information

```

I had previously looked at this in #741.

hwright:
I would also like to see this upgrade happen.  What are the chances we could bump the MSRV to 1.88 so that this would be possible?

PR #773, opened 2026-09-21
chore(deps): Update h2 to 0.4.19

hwright:
Versions `<0.4.16` have a low-priority vulnerability which causes our `cargo deny` check to fail:
https://rustsec.org/advisories/RUSTSEC-2026-0258

This only updates our the `h2` dependency (and it's dependencies) via `cargo update -p h2`.


epage:
My question is why this matters? This is through a dev-dependency

hwright:
One of the failing diagnostics points to this dependency:
https://github.com/rust-cli/config-rs/actions/runs/35603696336/job/106345590495

While it may be trivial as a dev dependency, it seems like the kind of thing to clean up so that real warnings are not overlooked.

